PR #19, opened 2017-06-14
Browser compatibility

### LINKED ISSUE #20: Browser compatibility?

noveens:
Fixes #20 

Introduced a make end point which will automatically make a browser compatible JS dist under the browser folder by the name, ```bundle.js```.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ ifndef VERBOSE
 .SILENT:
 endif
 
-.PHONY: all
+.PHONY: all browser
 
 all:
 	sudo npm i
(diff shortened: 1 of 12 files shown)


noveens:
@PVince81 , 
Is this okay?

Makefile
```diff
@@ -39,9 +39,24 @@ jsdocs:
 	#Output the final documentation link
 	echo "To read the documentation, click here : file://"${DIR}"/jsdoc/ownCloud.html"
 
+browser:
+	sed -i "s/require('fs/require('fs-web/g" owncloud/*.js
```
PVince81: `sed` sounds a bit barbaric here. Is there any best practice here that could be used ?

PVince81:
I suggest to not check in any generated files. Add them to `.gitignore`.

Let me people generate the bundles with make if they use the git version.
Later on you'll provide these precompiled bundles as release tarballs.

owncloud/helperFunctions.js
```diff
@@ -768,4 +774,26 @@ helpers.prototype._getFileName = function(path) {
...
 
+helpers.prototype._cleanseJson = function(json) {
+	for (var key in json) {
+		var a = recursiveCleanse(json[key]);
+		json[key] = a;
+	}
+	return json;
+};
+
+function recursiveCleanse(json) {
```
PVince81: what does this do any why ?

owncloud/owncloud.js
```diff
@@ -76,6 +83,34 @@ ownCloud.prototype.login = function(username, password) {
 };
 
 /**
+ * Logs in to the specified ownCloud instance (Updates capabilities)
+ * @param 		{string} 			  instance 		URL of the OC instance
```
PVince81: weird indents here

PVince81:
otherwise this looks fine, I didn't test it 👍

owncloud/helperFunctions.js
noveens: That's just for parsing the XML body to a JSON object.

Makefile
noveens: Why so?
Plus, I found many other solutions for the same, but couldn't find any for webPack.

PR #34, opened 2017-06-22
Not removing the namespace from DAV response

### LINKED ISSUE #30: oc.files.list() should give array of fileInfo classes
Update swagger schema as well indicating the same.

noveens:
Fixes #30 

@PVince81

### DIFF
--- a/owncloud/fileInfo.js
+++ b/owncloud/fileInfo.js
@@ -13,7 +13,7 @@ function fileInfo(name, type, attr) {
 	this.fileInfo = {};
 
 	for (var key in attr) {
-		this.fileInfo[key.split(':')[1]] = attr[key];
+		this.fileInfo[key] = attr[key];
 	}
 }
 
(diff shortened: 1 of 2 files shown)


owncloud/fileInfo.js
```diff
@@ -13,7 +13,7 @@ function fileInfo(name, type, attr) {
 	this.fileInfo = {};
 
 	for (var key in attr) {
-		this.fileInfo[key.split(':')[1]] = attr[key];
+		this.fileInfo[key] = attr[key];
```
PVince81: I'm surprised that everything else still works ? If you keep the namespace then you need to adjust your `FileInfo` class to parse the values from the array using the namespace notation

noveens: Yeah, sorry.
Successive PR: https://github.com/noveens/js-owncloud-client/pull/36

PR #48, opened 2018-03-21
Fix travis

### LINKED ISSUE #47: Travis seems to fail

DeepDiver1975:
fixes #47

### DIFF
--- a/owncloud/appManagement.js
+++ b/owncloud/appManagement.js
@@ -47,15 +47,14 @@ apps.prototype.getApps = function() {
         Promise.all([allAppsP, allEnabledAppsP])
             .then(apps => {
                 var tree = parser.xml2js(apps[0].body);
-                var statuscode = parseInt(helpers._checkOCSstatusCode(tree));
-                if (statuscode === 999) {
+                if (parseInt(helpers._checkOCSstatusCode(tree)) === 999) {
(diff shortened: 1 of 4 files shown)


owncloud/appManagement.js
```diff
@@ -53,9 +53,9 @@ apps.prototype.getApps = function() {
                     return;
                 }
 
-                if ((!(data in parser.xml2js(apps[0].body).ocs)) || (!(data in parser.xml2js(apps[1].body).ocs))) {
-                    reject(parser.xml2js(apps[0].body).ocs);
-                }
+                // if ((!(data in parser.xml2js(apps[0].body).ocs)) || (!(data in parser.xml2js(apps[1].body).ocs))) {
```
PVince81: why?

DeepDiver1975: data is undefined - no idea where data is supposed to come from.

deeper insights are welcome ....

PVince81: maybe it's supposed to be double quoted: `"data"` as the block seems to try and detect wheter "ocs.data" exists.

a simpler way to write this would be `if (parser.xml2js(apps[0].body).ocs.data)`

PVince81: probably not double quoted though... it does seem to be the equivalent of PHP's `isset`.

I suggest rewriting to the latter form

PR #63, opened 2018-05-23
Return user on login

felixheidecke:
`oc.login(…)` now returns the user object instead of simply returning `true`

Referring to: https://github.com/owncloud/js-owncloud-client/issues/59

### DIFF
--- a/docs-swagger/api/swagger/swagger.yaml
+++ b/docs-swagger/api/swagger/swagger.yaml
@@ -87,7 +87,16 @@ paths:
         "success":
           description: Login attemot was successful
           schema:
-            type: boolean
+      type: object
+      example: {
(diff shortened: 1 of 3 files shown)


PVince81:
please also adjust the unit tests

PVince81:
tests are failing because `'true'` is not `true`, didn't you fix this elsewhere?

felixheidecke:
@PVince81 I tried fixing it in https://github.com/owncloud/js-owncloud-client/pull/64 but local testing always fails on my Mac and Linux machine. Oshan didn't have any luck as well so I postponed.

Have you tried checking out the master and run `make test` ?

DeepDiver1975:
implemented in #100

PR #66, opened 2018-05-27
Get file as binary data

ittus:
## Description

Default encoding setting of [requests library](https://github.com/request/request) is `utf-8`. If we download `pdf` or `mp4` file, file will be corrupted.

> encoding - encoding to be used on setEncoding of response data. If null, the body is returned as a Buffer. Anything else (including the default value of undefined) will be passed as the encoding parameter to toString() (meaning this is effectively utf8 by default). (Note: if you expect binary data, you should set encoding: null.)

-> Set encoding to `null` will solve this problem

I don't know why there are 2 failed test cases. Please give me some clues to fix it.

## Relevant issues:
- https://github.com/owncloud/js-owncloud-client/issues/45
- https://stackoverflow.com/q/50546704/2303610

### DIFF
--- a/owncloud/helperFunctions.js
+++ b/owncloud/helperFunctions.js
@@ -350,7 +350,8 @@ helpers.prototype._get = function(url) {
     var options = {
         url: url,
         method: 'GET',
-        headers: headers
+        headers: headers,
+        encoding: null
     };
 
     return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


DeepDiver1975:
THX a lot for your effort on this @ittus - meanwhile we change the library rather dramatically.

I'm closing this and will add a test asap ... THX

PR #108, opened 2018-11-05
Missing return after reject

DeepDiver1975:


PVince81:
no test seems affected :-/

PR #110, opened 2018-11-09
Use davclient.js for DAV requests to owncloud

DeepDiver1975:
fixes #110

### DIFF
--- a/owncloud/fileManagement.js
+++ b/owncloud/fileManagement.js
@@ -3,7 +3,10 @@
 //////////////////////////////////////
 
 var Promise = require('promise');
+var dav = require('davclient.js');
+var fileInfo = require('./fileInfo.js');
 var helpers;
+var davClient;
 
 /**
  * @class files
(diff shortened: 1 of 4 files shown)


owncloud/fileManagement.js
```diff
@@ -236,12 +280,88 @@ files.prototype.move = function(source, target) {
...
+    if (response.propStat.length === 0 || response.propStat[0].status !== 'HTTP/1.1 200 OK') {
+        return null;
+    }
+
+    var props = response.propStat[0].properties;
+    let fileType = 'file';
+    var resType = props['{DAV:}resourcetype'];
+    if (resType) {
+        var xmlvalue = resType[0];
+        if (xmlvalue.namespaceURI === 'DAV:' && xmlvalue.nodeName.split(':')[1] === 'collection') {
```
PVince81: I hope this works in IE 11. should be possible to run the karma tests at least once in a local IE 11 instance once done

DeepDiver1975: IE11 is a topic on it's own - afaik we need some polyfills for it anyway .... will create topic ... THX

PR #112, opened 2018-11-13
Feature/serv3r cypress

scramb:
i created testFolder + subdir & "in dir.txt" manually test is working need to be adjusted asap fully integrated


cypress/integration/sample_spec.js
```diff
@@ -1,33 +1,40 @@
...
 
+
     beforeEach(function (done) {
         oc = new ownCloud(config.owncloudURL);
-        oc.login(config.username, config.password).then(status => {
-            expect(status).toEqual({id: 'admin', 'display-name': 'admin', email: {}});
-            done();
-        }).catch(error => {
+        ocLogin = oc.login(config.username, config.password).then(res => {
+          cy.wrap(res).its('id').should('eq', config.username);
```
DeepDiver1975: so we need yet another assert syntax - I'm getting more and more concerned if this is the right way ...

owncloud/test/config.sample.json
```diff
@@ -1,5 +1,5 @@
 {
-    "owncloudURL": "http://127.0.0.1/",
+    "owncloudURL": "http://127.0.0.1:8888/",
```
DeepDiver1975: why did you set the owncloud url to the same as the cypress url? I doubt this is working

scramb: Cypress baseUrl is just the address he sends requests to. I set it manually here since I did not look into changing this on-the-fly. We could handle this by calling cypress.config('baseUrl', the.url.owncloud.core.runs.on).

cypress/integration/sample_spec.js
scramb: The Lib is writen in a way cypress does have trouble with since it's never run in a browser and without cors since there is no browseable site. This is a workaround and should be more like 
`
cy.visit(baseUrl);
cy.request('login', config.username, config.password)
`
im really not sure if this is going to work for the lib

DeepDiver1975: I sorted cors already out - look at my pr - the setupoc script is setting the domain of cypress.
this is by it should stay at port 5555

cy.visit will not work - we have no page to visit.

this unit test is modelled after https://github.com/cypress-io/cypress-example-recipes/blob/master/examples/unit-testing__application-code/cypress/integration/unit_test_application_code_spec.js

this seems to be a valid apporach


never the less .... I'm not yet fully convinced ....

scramb: > I sorted cors already out - look at my pr - the setupoc script is setting the domain of cypress.
> this is by it should stay at port 5555

this could be true, but the baseUrl does tell where to send requests at what should be the owncloud running on :8888 , right? since we're not using cy.request or cy.visit this value is not used once. We need to specify the port Option as 55555 not the baseUrl.

scramb: actually cypress doesn't open a browser since we're never using cy.visit i'll prepare a PR for phoenix for how it's supposed to work. To be honest i'm kinda sure cypress is not the right way for library testing.

PR #114, opened 2018-11-16
Add codecoverage

DeepDiver1975:
code coverage is now generated and pushed to codecov.io

### DIFF
--- a/tests/karma.js
+++ b/tests/karma.js
@@ -1,0 +1,7 @@
+/* eslint-disable new-cap */
+var owncloud = require('../owncloud/index.js')
+var oc = new owncloud()
+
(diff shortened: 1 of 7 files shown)


tests/karma.js
```diff
@@ -0,0 +1,7 @@
+/* eslint-disable new-cap */
+var owncloud = require('../owncloud/index.js')
+var oc = new owncloud()
+
+window.oc = oc
```
DeepDiver1975: this has to go away - tests need to be written in es6 and transpiled

DeepDiver1975:
> unless you need minimal time to rewrite this in ES6...

I'll do this in a follow up pr - THX

PR #129, opened 2018-11-23
Adding oc.files.star

DeepDiver1975:


src/fileInfo.js
```diff
@@ -86,12 +86,20 @@ fileInfo.prototype.getContentType = function () {
...
  * Gets last modified time of file/folder
- * @returns {integer}   Last modified time of file/folder
+ * @returns {Date}   Last modified time of file/folder
  */
 fileInfo.prototype.getLastModified = function () {
   return new Date(this.fileInfo['{DAV:}getlastmodified'])
 }
 
+/**
+ * Gets last modified time of file/folder
```
PVince81: gets arbitrary property ?

src/fileManagement.js
```diff
@@ -218,6 +217,7 @@ files.prototype.delete = function (path) {
 /**
  * Returns the file info for the given remote file
  * @param   {string}  path          path of the file/folder at OC instance
+ * @param   {array}   properties    WebDAV properties
```
PVince81: is this an array of the property names `{Array.<string>}` ? or a hash map of property name to values `{Object.<string, string>}` ?

src/fileManagement.js
```diff
@@ -318,4 +318,35 @@ files.prototype.copy = function (source, target) {
...
 }
 
+/**
+ * Mark a remote file or directory as favorite
+ * @param   {string} path       path of file/folder
+ * @param   {boolean} value     Star or un-star the file/folder
+ * @returns {Promise.<status>}  boolean: whether the operation was successful
+ * @returns {Promise.<error>}   string: error message, if any.
+ */
+files.prototype.star = function (path, value) {
```
PVince81: we usually call it "favorite" everywhere in the core code, never "star". it's only the icon that is a star.
might be better to keep it consistent to avoid future confusion

src/fileInfo.js
```diff
@@ -86,12 +86,20 @@ fileInfo.prototype.getContentType = function () {
...
- * @returns {integer}   Last modified time of file/folder
+ * @returns {Date}   Last modified time of file/folder
  */
 fileInfo.prototype.getLastModified = function () {
   return new Date(this.fileInfo['{DAV:}getlastmodified'])
 }
 
+/**
+ * Gets arbitrary property
+ * @returns {string}   Value of the property
```
PVince81: and `@param {string} property Name of the property for which to retrieve the value`

also in general I wonder whether we had JSDoc setup?

DeepDiver1975: https://owncloud.github.io/js-owncloud-client/

PR #133, opened 2018-11-27
FilesVersions: restore and get contents of versions has been added

### LINKED ISSUE #93: Add support for querying versions of a file
- [x] REQUIRES https://github.com/owncloud/js-owncloud-client/issues/25 to be able to query file id
- [x] oc.fileVersions.listVersions(fileId)
- [ ] oc.fileVersions.getVersionContents(fileId, versionId)
- [ ] oc.fileVersions.restoreVersion(fileId)
- [ ] have an object VersionInfo that extends FileInfo (or whatever makes sense):
    - fileId
    - versionId
    - mtime (timestamp when the version was created)
    - size
    - previewURl

DeepDiver1975:
closes #93

### DIFF
--- a/src/fileVersionManagement.js
+++ b/src/fileVersionManagement.js
@@ -9,21 +9,11 @@ var davClient
  * <b><i> The FilesVersions class, has all the methods for your ownCloud files version management.</i></b><br><br>
  * Supported Methods are:
  * <ul>
- *  <li><b>Files Management</b>
+ *  <li><b>Files Versions Management</b>
  *      <ul>
- *          <li>list</li>
- *          <li>getFileContents</li>
- *          <li>putFileContents</li>
- *          <li>mkdir</li>
- *          <li>createFolder</li>
- *          <li>delete</li>
- *          <li>fileInfo</li>
- *          <li>getFile</li>
- *          <li>getDirectoryAsZip</li>
- *          <li>putFile</li>
- *          <li>putDirectory</li>
- *          <li>move</li>
- *          <li>copy</li>
+ *          <li>listVersions</li>
+ *          <li>getFileVersionContents</li>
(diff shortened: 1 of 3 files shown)


src/fileVersionManagement.js
```diff
@@ -67,4 +57,61 @@ FilesVersions.prototype.listVersions = function (fileId) {
   })
 }
 
+FilesVersions.prototype.getFileVersionContents = function (fileId, versionId) {
```
PVince81: please add JSDoc

src/fileVersionManagement.js
```diff
@@ -67,4 +57,61 @@ FilesVersions.prototype.listVersions = function (fileId) {
...
+        var err = helpers._parseDAVerror(body)
+        reject(err)
+      }
+    }).catch(error => {
+      reject(error)
+    })
+  })
+}
+
+FilesVersions.prototype.restoreFileVersion = (fileId, versionId, targetPath) => {
```
PVince81: please add JSDoc

src/fileVersionManagement.js
```diff
@@ -67,4 +57,61 @@ FilesVersions.prototype.listVersions = function (fileId) {
...
+  return new Promise((resolve, reject) => {
+    if (!helpers.getAuthorization()) {
```
PVince81: oh, we have a stateful global helper class ?

can we rather encapsulate this into a class and inject it in the constructor ? (separate PR)

PVince81:
well, fix CI and the JSDoc

src/fileVersionManagement.js
DeepDiver1975: The helpers Class is injected in the ctor

PR #135, opened 2018-12-04
Add OwnCloud.requests.ocs

DeepDiver1975:


PVince81:
@DeepDiver1975 mind adding some context in the ticket ? I see OCS but am not sure what this PR supposed to do

src/helperFunctions.js
```diff
@@ -163,11 +165,13 @@ helpers.prototype._makeOCSrequest = function (method, service, action, data) {
   if (service) {
     slash = '/'
   }
-
   var path = this.OCS_BASEPATH + service + slash + action
+  if (options.useV2) {
```
PVince81: should we use V2 by default ?

DeepDiver1975: I will step by step migrate this. but the current implementation of makeOCSRequest is relying on v1 to be used. I'll drop the v2 part ... it did not work.

THX for a second pair of eyes - I missed that

PR #141, opened 2018-12-17
Send X-Request-ID with every request ...

DeepDiver1975:
refs #140 

blocked by https://github.com/owncloud/core/pull/33909

### DIFF
--- a/src/helperFunctions.js
+++ b/src/helperFunctions.js
@@ -112,6 +112,7 @@ helpers.prototype._updateCapabilities = function () {
 
         self._capabilities = body.capabilities
         self._version = body.version.string + '-' + body.version.edition
+        self._versionNumber = body.version.major + '.' + body.version.minor + '.' + body.version.micro
 
         resolve(self._capabilities)
       }).catch(error => {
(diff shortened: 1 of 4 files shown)


PVince81:
how to fix CI ? you'd need the docker in Travis to point at stable10

DeepDiver1975:
No idea how to handle this. We should only send the Header with 10.1 and higher

PVince81:
You can detect version with capabilities I believe

DeepDiver1975:
So the capabilities requests never sends the header then .... Sounds reasonable

DeepDiver1975:
@PVince81 mind reviewing this one again? THX

src/helperFunctions.js
```diff
@@ -140,6 +142,25 @@ helpers.prototype._updateCurrentUser = function () {
   })
 }
 
+helpers.prototype.buildHeaders = function () {
+  var headers = {
+    authorization: this._authHeader,
+    'OCS-APIREQUEST': true
+  }
+  if (this.atLeastVersion('10.1.0')) {
```
PVince81: hmm ok so no capability then ?

PR #159, opened 2019-02-06
Add If-Match and If-Not-Match headers when putting files

DeepDiver1975:


DeepDiver1975:
@PVince81 ready for review - THX

.travis.yml
```diff
@@ -1,27 +1,35 @@
...
+  only:
+  - master
+
+env:
+  - OC_HASH: bb8b8878b40edaeeba831663275d9815272cdea2
+
 before_install:
-  - docker pull owncloud:$OC_VERSION
-  - DOCKER_ID=$(docker run -d -p 8888:80 owncloud:$OC_VERSION)
+  - wget https://github.com/owncloud/core/archive/$OC_HASH.zip
```
PVince81: will we use dailies in the future ?

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM owncloud/base:bionic
+
+LABEL maintainer="ownCloud DevOps <devops@owncloud.com>" \
+  org.label-schema.name="ownCloud Server" \
+  org.label-schema.vendor="ownCloud GmbH" \
+  org.label-schema.schema-version="1.0"
+
+ADD owncloud-*.tar.bz2 /var/www/
+
+RUN find /var/www/owncloud \( \! -user www-data -o \! -group root \) -print0 | xargs -r -0 chown www-data:root && \
```
PVince81: don't we have an official docker that already does this ?

.travis.yml
DeepDiver1975: yes

tests/appsTest.js
```diff
@@ -1,6 +1,6 @@
 describe('Main: Currently testing apps management,', function () {
 // CURRENT TIME
-  var timeRightNow = new Date().getTime()
+  var timeRightNow = Math.random().toString(36).substr(2, 9)
```
PVince81: better use hardcoded values for deterministic results

Dockerfile
DeepDiver1975: Not needed - will drop this file ...

tests/appsTest.js
DeepDiver1975: The idea is to add a suffix on each test run so that files and folders don't collide when rerunning the test suite.

DeepDiver1975:
> raise ticket for using official docker ?

I'm using an official docker image - the Dockerfile is a left-over ....

tests/appsTest.js
PVince81: ah ok

PR #166, opened 2019-02-18
Test with daily master and daily stable10

DeepDiver1975:


DeepDiver1975:
Needs to be re-executed after Feb19th ...

Issue #184, opened 2019-04-05
Push js-owncloud-client to npmjs via drone - requires to move the project from travis to drone


DeepDiver1975:

DeepDiver1975:
refs #186

DeepDiver1975:
finally finished with #201

PR #199, opened 2019-04-23
Setup babel and browserslist

DeepDiver1975:
refs https://github.com/owncloud/owncloud-sdk/issues/184


DeepDiver1975:
reverting - this is not working .... :man_shrugging:

PR #217, opened 2019-05-06
Add system tags apis and filtering

DeepDiver1975:


tests/filesTest.js
```diff
@@ -205,6 +205,28 @@ describe('Main: Currently testing files management,', function () {
     })
   })
 
+  it('checking method: favorite filter', function (done) {
+    const newFile = testFolder + '/' + 'file.txt'
+
+    oc.files.putFileContents(newFile, testContent).then(status => {
+      expect(typeof status).toBe('object')
+      return oc.files.favorite(newFile)
+    }).then(status2 => {
```
PVince81: wondering if we should invest more in async-await in the future as I remember it could make this kind of chain more readable

src/systemTags.js
```diff
@@ -0,0 +1,71 @@
...
+    })
+  }
+
+  tagFile (fileId, tagId) {
+    if (!this.helpers.getAuthorization()) {
+      return Promise.reject('Please specify an authorization first.')
+    }
+    const path = 'systemtags-relations/files/' + fileId + '/' + tagId
+
+    return this.davClient.request('PUT',
```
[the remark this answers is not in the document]
DeepDiver1975: yes - eslint is in place ... see the .eslintrc.json

tests/filesTest.js
```diff
@@ -205,6 +205,28 @@ describe('Main: Currently testing files management,', function () {
     })
   })
 
+  it('checking method: favorite filter', function (done) {
```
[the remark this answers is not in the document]
DeepDiver1975: needs to be applied to all it() ... one day ...

but yes - I agree ...

tests/filesTest.js
DeepDiver1975: indeed worth a look :+1:

DeepDiver1975:
@PVince81 please re-review - THX

src/systemTags.js
```diff
@@ -0,0 +1,106 @@
+var Promise = require('promise')
+const dav = require('davclient.js')
+
+/**
+ * @class Files
+ * @classdesc
+ * <b><i> The Files class, has all the methods for your ownCloud files management.</i></b><br><br>
+ * Supported Methods are:
+ * <ul>
+ *  <li><b>Files Management</b>
```
PVince81: please adjust

PVince81: I guess we could just remove those, I expect we should be able to generate proper API docs from the JSDoc blocks

PR #313, opened 2019-08-08
Add X-Requested-With header - this is used by the server to determine…

DeepDiver1975:
… if the request is made by the user or via an ajax call

see http://www.checkheaders.com/headers/X-Requested-With/
see https://github.com/owncloud/core/blob/3bff5c2aca12499972f285e0cf2ea467a473c6b5/apps/dav/lib/Connector/Sabre/Auth.php#L240-L245

### DIFF
--- a/src/helperFunctions.js
+++ b/src/helperFunctions.js
@@ -119,7 +119,8 @@ class helpers {
 
   buildHeaders (withAuthHeader = true) {
     let headers = {
-      'OCS-APIREQUEST': true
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
this will only work after https://github.com/owncloud/core/pull/35932/commits/87575f33764b86cd2168a78cf679ee7ec56a2b94 has been merged ..... need to find a less breaking solution ... maybe by adding a config option .....

PR #345, opened 2019-09-12
Add json format to capabilities request

LukasHirt:
## Motivation
When we got response in XML format the booleans were translated into string which caused desync between oc and reva.

@refs


tests/requestsOcsTest.js
```diff
@@ -33,7 +33,7 @@ describe('Main: Currently testing low level OCS', function () {
   it('checking : capabilities', function (done) {
     oc.requests.ocs({
       service: 'cloud',
-      action: 'capabilities'
+      action: 'capabilities?format=json'
```
DeepDiver1975: oc.requests.ocs is already using json format - this is not needed ....

PR #350, opened 2019-09-23
Use fetch API only ....

DeepDiver1975:
:construction:


PVince81:
does it mean we also switch away from davclient.js ? or need to add fetch support to davclient.js as well ? (https://github.com/owncloud/davclient.js/issues/36)

PR #366, opened 2019-10-14
Add upload progress + headers

PVince81:
Requires https://github.com/owncloud/davclient.js/pull/39/

### DIFF
--- a/src/fileManagement.js
+++ b/src/fileManagement.js
@@ -141,6 +141,10 @@ class Files {
    * @param   {string} path       path of the file at OC instance
    * @param   {string} content    content to be put
    * @param   {Object} options
+   * @param   {Object} [options.headers] optional extra headers
+   * @param   {boolean} [options.overwrite] whether to force-overwrite the target
+   * @param   {String} [options.previousEntityTag] previous entity tag to avoid concurrent overwrites
+   * @param   {Function} options.onProgress progress callback
    * @returns {Promise.<status>}  boolean: whether the operation was successful
    * @returns {Promise.<error>}   string: error message, if any.
    */
@@ -149,17 +153,22 @@ class Files {
       return Promise.reject('Please specify an authorization first.')
     }
     options = options || []
-    const headers = this.helpers.buildHeaders()
+    const headers = Object.assign({}, this.helpers.buildHeaders(), options.headers)
     const previousEntityTag = options.previousEntityTag || false
     if (previousEntityTag) {
       // will ensure that no other client uploaded a different version meanwhile
       headers['If-Match'] = previousEntityTag
-    } else {
+    } else if (!options.overwrite) {
       // will trigger 412 precondition failed if a file already exists
       headers['If-None-Match'] = '*'
     }
 
-    return this.davClient.request('PUT', this.helpers._buildFullWebDAVPath(path), headers, content).then(result => {
+    let requestOptions = {}
+    if (options.onProgress) {
+      requestOptions.onProgress = options.onProgress
+    }
+
+    return this.davClient.request('PUT', this.helpers._buildFullWebDAVPath(path), headers, content, null, requestOptions).then(result => {
       if ([200, 201, 204, 207].indexOf(result.status) > -1) {
         return Promise.resolve({
           'ETag': result.xhr.getResponseHeader('etag'),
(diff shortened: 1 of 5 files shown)


PVince81:
- [ ] TODO: apply the same for public links

or refactor the whole thing to use only one code path for uploads...

package.json
```diff
@@ -31,8 +31,10 @@
     "lint-fix": "eslint tests/**/*.js tests/*.js src/**/*.js src/*.js --color --global requirejs --global require --fix",
     "test": "karma start --single-run --browsers ChromeHeadless",
     "test-drone": "karma start --single-run --browsers ChromeHeadlessNoSandbox",
+    "prepare": "npm run build:dev",
```
DeepDiver1975: Is that needed within the final PR in here? I'm aware of the local dev reasons for this - I'm wondering about sideeffects on the real build/publish - THX

PVince81: I hope we can keep it including the source map thing, especially if we're going to do more development on the SDK in the near future.

But I have the same worry like yours. I'll research this

PVince81:
- [ ] research whether "prepare" is skipped for production builds

PVince81:
According to https://docs.npmjs.com/misc/scripts it seems "prepare" is only using when publishing packages and when installing git dependencies.

But then it means that if I put the source map there (the dev build) it would also package that in the production version.

Doesn't seem like a good solution so I'm fine removing these bits before merging.

PVince81:
ref where I found out about npm prepare: https://stackoverflow.com/questions/48287776/automatically-build-npm-module-on-install-from-github

package.json
DeepDiver1975: We could also skip webpack totally - I did add webpack to get transpiling in place.
With the knowledge of today: any consuming project can transpile any dependency to it's own needs.

the sdk can be shippes as pure sources ... which would also make debugging on the consuming project (e.g. phoenix) easier

PVince81: if we use ES6, would transpiling be needed for running the tests in the browser ?

DeepDiver1975: tests have anyhow their own setup/config - this is unrelated to the way we distribute the package

DeepDiver1975:
In order to get this rolling I suggest to split the PR: upload progress and build setup

PVince81:
added code for public link
added onProgress assertion in the tests (and used the opportunity to use async/await there)

PVince81:
finally got tests passing.

next up: cleanup

PVince81:
@DeepDiver1975 I've removed the "dev mode" stuff, but moved it to a separate branch for later reconsideration.

@DeepDiver1975 please review, this is ready to ship

DeepDiver1975:
@PVince81 there is something wrong in the area of public files - see ci ....

PVince81:
it is possible that for small files `onProgress` is never called... the spec doesn't tell how often a browser is supposed to call it, it just says "periodically".

maybe for this one test we should mock the davclient...

PVince81:
all tests pass for me locally :frowning:

PVince81:
I've found this, so it could be timing related: https://stackoverflow.com/a/5543745

I'll see if it's easier to use a mock instead

PVince81:
- [ ] create new folder "unit" for unit tests using mocks when needed, to make sure there is a clear separation

PVince81:
seems Sinon JS have changed their APIs and I can't find the progress object https://sinonjs.org/releases/latest/fake-xhr-and-server/
also no luck with Jasmine's ajax XHR mock which doesn't have the "upload" attribute https://github.com/jasmine/jasmine-ajax/issues/109

PVince81:
as an alternative attempt, I tried uploading 1 MB+ content but the progress events still don't fire, despite the stack overflow article there. I wonder if this is related to ChromeHeadless that progress events are not fired

PVince81:
another hope here that would bring us closer to how browsers actually upload: https://stackoverflow.com/a/43767079

- [ ] try uploading with FormData or other methods like binary blob, as they are more likely to trigger progress events

PVince81:
Uploading with FormData produces extra contents, but not usable.

- [x] try with Blob

Unfortunately uploading a Blob directly also does not trigger progress events :rage:

PVince81:
Big facepalm!

The yarn.lock was still pointing at the old davclient.js that had no onProgress event.
I didn't think of going to debug that deep...

Updated now, tests should pass!

PVince81:
@DeepDiver1975 this is ready for re-review and merging!

PR #437, opened 2020-03-26
Get fileInfo of publicly shared resource

LukasHirt:


src/publicLinkAccess.js
```diff
@@ -251,5 +251,23 @@ class PublicFiles {
       return Promise.reject(this.helpers.buildHttpErrorFromDavResponse(result.status, result.body))
     })
   }
+
+  /**
+   * Returns the file info for the given public resource
+   * @param   {string} token public share token - may contain the path as well
+   * @param   {string|null} password public link's password
+   * @param   {Object.<string, string>} properties WebDAV properties
```
kulmann: According to `list` docblock this should be of type `array` (`* @param {array}       properties`)

src/publicLinkAccess.js
```diff
@@ -251,5 +251,23 @@ class PublicFiles {
       return Promise.reject(this.helpers.buildHttpErrorFromDavResponse(result.status, result.body))
     })
   }
+
+  /**
+   * Returns the file info for the given public resource
+   * @param   {string} token public share token - may contain the path as well
```
kulmann: Looking at `list` it looks like only the token may be provided, not the path.

DeepDiver1975:
Is it necessary to add more code? list (which is actually PROPFIND) wiht depth 0 does exactly this.

LukasHirt:
> Is it necessary to add more code? list (which is actually PROPFIND) wiht depth 0 does exactly this.

My point was to make this easier for developers. Also, if some external developers create some extension using the sdk then it's easier when they'd see this function instead of having to explain they can do it with the list function.

src/publicLinkAccess.js
LukasHirt: Right - fixed.

src/publicLinkAccess.js
LukasHirt: It needs to contain also path otherwise we wouldn't be able to reach the requested file. Also, in the list, it's called actually `tokenAndPath`.

https://github.com/owncloud/owncloud-sdk/blob/42936db99adeec5dd40a4f79128db4c48283cf45/src/publicLinkAccess.js#L41

I can rename it to the same to be consistent.

LukasHirt: Ah, now I see where is the problem. `getFileUrl` is not used correctly. I'll fix that and adjust the code accordingly.

karma.conf.js
```diff
@@ -13,6 +13,15 @@ module.exports = function (config) {
     // base path that will be used to resolve all patterns (eg. files, exclude)
     basePath: '',
 
+    plugins: [
```
LukasHirt, line 16: Plugins need to be included otherwise they won't function properly.

karma.conf.js
```diff
@@ -67,7 +76,7 @@ module.exports = function (config) {
     // test results reporter to use
     // possible values: 'dots', 'progress'
     // available reporters: https://npmjs.org/browse/keyword/karma-reporter
-    reporters: ['dots', 'coverage-istanbul'],
+    reporters: ['progress', 'coverage-istanbul'],
```
LukasHirt, line 79: When running the tests `dots` are not telling us anything. `progress` seems to me as a much better option.

LukasHirt:
Strange failure in tests. It passes locally though. 😕

PR #440, opened 2020-04-03
Update dependencies and apply new eslint rules

DeepDiver1975:


src/helperFunctions.js
```diff
@@ -113,7 +113,7 @@ class helpers {
    * @returns {Promise.<error>}           string: error message, if any.
    */
   _updateCurrentUser () {
-    let self = this
+    const self = this
```
kulmann, line 116: I know that you didn't touch it, but `self` should not be necessary here, as `this` is automatically bound as context in lambda functions.

package.json
```diff
@@ -122,7 +122,7 @@
   },
   "lint-staged": {
     "*.js": [
-      "lint --fix",
+      "./node_modules/.bin/eslint --fix",
```
kulmann, line 125: why not use the `lint-fix` script?

DeepDiver1975: lint --fix as well as lint-fix are failing without output - no idea why.
this way it is working ... :man_shrugging:

src/helperFunctions.js
DeepDiver1975: Can we agree to change this in a follow up PR?

kulmann: 👍

Issue #458, opened 2020-04-21
Expose TUS capability of a resource


PVince81:
When a Webdav resource supports TUS, any request done to it (PROPFIND, OPTIONS) will return a `Tus-Version: 1.0.0` header whenever the backend storage supports TUS (see https://github.com/cs3org/reva/pull/661 and https://github.com/owncloud/phoenix/pull/3345)

The problem is that none of the relevant calls are exposing the headers to the caller.
For example here the `client.files.list()` call is directly returning an array.
And for a single file `client.files.fileInfo()` also returns the file info directly.
There is no room to add response headers or a new attribute `tusSupported` there without breaking API compatibility.

Here are other possible exotic approaches:
- Approach 0: break API compatibility and change the response format
- Approach 1: add a new attribute `options.returnAsObject` which changes the response format to the new thing
- Approach 2: duplicate the methods for which we change the return types and deprecate the olds ones
- Approach 3 (dirty hack): the API caller wraps the XHR object before the call and captures the response headers there
- Approach 4 (inelegant): have the SDK expose a "getLastHeaders()" function somewhere
- Approach 5: have an additional API call that does an OPTIONS call to an endpoint to return capabilities and headers. Drawback: an additional call.
- Approach 6 (dirty hack): add the headers array as an attribute of the array returned by `list()`

Thoughts ? @DeepDiver1975 @butonic @LukasHirt

PVince81:
https://github.com/owncloud/owncloud-sdk/blob/master/src/fileManagement.js#L58

PVince81:
PR here: https://github.com/owncloud/owncloud-sdk/pull/461

PR #461, opened 2020-04-22
Add TUS support information on FileInfo object

PVince81:
Whenever a list() or fileInfo() call response contains TUS headers,
return these through the fileInfo.getTusSupport() method.

If no TUS headers are found, that method returns null.

Fixes https://github.com/owncloud/owncloud-sdk/issues/458

- [x] unit tests
- [x] test against actual backend

### DIFF
--- a/tests/filesTest.js
+++ b/tests/filesTest.js
@@ -1,8 +1,10 @@
 describe('Main: Currently testing files management,', function () {
   // CURRENT TIME
   const timeRightNow = Math.random().toString(36).substr(2, 9)
+  const FileInfo = require('../src/fileInfo')
   const OwnCloud = require('../src/owncloud')
   const config = require('./config/config.json')
+  const sinon = require('sinon')
 
   // LIBRARY INSTANCE
   let oc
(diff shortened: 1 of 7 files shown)


PVince81:
I've re-read the spec and apparently only "Tus-Version" MUST be set and not the others.

- [ ] rely on "Tus-Version" instead of "Tus-Resumable" to detect if TUS is set at all

PVince81:
adjusted, please review

tests/filesTest.js
```diff
@@ -523,4 +525,85 @@ describe('Main: Currently testing files management,', function () {
...
+    afterEach(function () {
+      parseBodyStub.restore()
+      xhr.restore()
+    })
+
+    it('returns TUS support information when TUS headers are set for a list call', function (done) {
+      const promise = oc.files.list('')
+      promise.then(entries => {
+        const tusSupport = entries[0].getTusSupport()
+        expect(tusSupport.resumable).toEqual('1.0.0')
```
LukasHirt, line 554: > I've re-read the spec and apparently only "Tus-Version" MUST be set and not the others.

Should this line still be here then or removed?

PVince81: I think we can keep it as I want to check whether the header is parsed when it's present (the test is only using a mock anyway)

Issue #462, opened 2020-04-24
Refused to get unsafe header "tus-version"


PVince81:
Since the addition of [TUS](#3345) we are now checking the TUS capability from the backend by reading the "tus-version" header.

If that one is not available, we should fail silently.

When using OC 10 as backend we get the error `Refused to get unsafe header "tus-version"` because it's not part of CORS.

I suggest we add a try/catch block in the SDK when parsing responses

PVince81:
Attempt 1: trying `try/catch` around the get call => :boom: 

Attempt 2: trying to read the list of exposed headers before getting the unsafe header:
<img width="339" alt="image" src="https://user-images.githubusercontent.com/277525/80197445-f1144600-861e-11ea-9ec8-edb2c6a4042f.png">

:exploding_head:

PVince81:
Attempt 3: using `getAllResponseHeaders` to check beforehand whether the tus header is there without tripping over that error

PVince81:
attempt 3 successful, PR here: https://github.com/owncloud/owncloud-sdk/pull/463

PR #463, opened 2020-04-24
Prevent logging unsafe header error

PVince81:
Backend that don't support TUS will also not expose TUS headers through
CORS. To prevent the uncatchable "refused to get unsafe header", this
commit adds a workaround with a preliminary check using
"getAllResponseHeaders" to find out if the header is exposed by CORS
before reading it.

Fixes https://github.com/owncloud/owncloud-sdk/issues/462


PVince81:
- [ ] fix tests to also mock the getAllResponseHeaders call

PVince81:
I've fixed the tests by lower casing the "all headers" call, since the mocker from sinon did not lower case by default. It might also increase compatibility in case some browsers do not lower case by default either.

PR #475, opened 2020-05-20
Fetch settings values from settings service

kulmann:
This PR introduces a new `settings` class, which is responsible for fetching settings values from the settings service. It does so by using the generated JavaScript client (`src/settingsClient.js`). So please don't review `src/settingsClient.js`, as it's generated.

As said generated client relies on Axios, this PR also introduces Axios. The bearer token - if present - is injected in `src/helperFunctions.js`, so that axios requests can be made everywhere else in the code as well.

The function `getSettingsValues` from `src/settings.js` will be used in Phoenix. I will link a Phoenix PR here, soon. The code is supposed to fall back to defaults, if the settings service is unavailable. We'll have to discuss, if those defaults are supposed to live in the SDK or in Phoenix. But in my opinion that can be solved later on.


kulmann:
At the moment the generated JS client for retrieving settings is using POST requests internally, for everything. We are working on changing this to something more RESTful.

kulmann:
Are you ok with moving the generated settings client into a `vendor` dir? Not so happy with it living the `src` folder...

LukasHirt:
> Are you ok with moving the generated settings client into a vendor dir? Not so happy with it living in the src folder...

Yeah, would be nice to make it clear this way that it's not something developers should edit.

src/settings.js
```diff
@@ -0,0 +1,46 @@
+const SettingsClient = require('./settingsClient')
+const Promise = require('promise')
+
+/**
+ * @class SettingsValues
+ * @classdesc
+ * <b><i> The SettingsValues class provides access to all settings values of the (most of the time authenticated) user.</i></b>
+ *
+ * @author Noveen Sachdeva
```
PVince81, line 9: fix author ?

src/settings.js
```diff
@@ -0,0 +1,46 @@
...
+    this.helpers = helpers
+  }
+
+  /**
+   * Gets all settings values for the given account uuid. If no uuid is provided, settings values
+   * for the authenticated user will be fetched.
+   * @returns {Promise.<settings>} array: all available settings values
+   * @returns {Promise.<error>}    string: error message, if any.
+   */
+  async getSettingsValues (accountUuid = 'me') {
```
PVince81, line 24: is "me" the default ? if yes please add in the JSDoc that "me" will be used as default and also what it means (probably the user's profile?)

src/settings.js
```diff
@@ -0,0 +1,46 @@
...
+          identifier: {
+            account_uuid: accountUuid
+          }
+        }
+      })
+      if (response.status === 201) {
+        return Promise.resolve(response.data.settingsValues)
+      }
+    } catch (error) {
+      // fall through to returning defaults
```
PVince81, line 39: are all errors valid to fall through ? or just 404 ?

src/settings.js
```diff
@@ -0,0 +1,46 @@
...
+ * @param {object}  helperFile  instance of the helpers class
+ */
+class SettingsValues {
+  constructor (helpers) {
+    this.helpers = helpers
+  }
+
+  /**
+   * Gets all settings values for the given account uuid. If no uuid is provided, settings values
+   * for the authenticated user will be fetched.
```
PVince81, line 20: also add that if no settings service is reachable, no error is returned but an empty hash.

I wonder if we should rather still return an error (like "service unavailable or so") and have the caller decide what to do, because the caller might know whether the settings service is configured or not ?

src/settingsClient.js
```diff
@@ -0,0 +1,305 @@
...
+import axios from 'axios'
+import qs from 'qs'
+let domain = ''
+export const getDomain = () => {
+  return domain
+}
+export const setDomain = ($domain) => {
+  domain = $domain
+}
+export const request = (method, url, body, queryParameters, form, config) => {
```
PVince81, line 11: I guess one can pass in headers in "config" ?

src/settings.js
kulmann: Yes, `me` is the default. Added param and description to the docblock.

src/settings.js
kulmann: atm it's a 502, not 404. changed to Promise.reject on anything other than 502.

src/settings.js
kulmann: the javascript console always shows failed requests, so the service not being available will be visible in the JS console. can't suppress that.

src/settingsClient.js
kulmann: FYI, bearer token is injected automatically here: https://github.com/owncloud/owncloud-sdk/blob/440966fc7071ea27ee91195bfcf3202e6f7d4851/src/helperFunctions.js#L61

src/settings.js
```diff
@@ -0,0 +1,54 @@
...
+          }
+        }
+      })
+      if (response.status === 201) {
+        return Promise.resolve(response.data.settingsValues)
+      }
+    } catch (error) {
+      // fail on anything except settings service being unavailable
+      // TODO: if this changes to GET requests in the future, this will be 404 instead of 502
+      if (error.response.status !== 502) {
```
PVince81, line 45: now thinking, there might also be a risk that if the service is just down (but configured) that Phoenix would revert to default values instead of showing an error. we can consider this case in a separate issue

vendor/settingsClient.js
```diff
@@ -0,0 +1,305 @@
+/* eslint-disable */
+import axios from 'axios'
```
PVince81, line 2: if this is a vendor thing with no npm page, where does this come from ? how to update it ?

kulmann: added a comment

PR #479, opened 2020-05-25
Also ignore 404 for settings service, not just 502.

kulmann:
Had to add 404 to the ignored error codes. Was only ignoring 502 before.

### DIFF
--- a/src/settings.js
+++ b/src/settings.js
@@ -41,8 +41,7 @@ class SettingsValues {
       }
     } catch (error) {
       // fail on anything except settings service being unavailable
-      // TODO: if this changes to GET requests in the future, this will be 404 instead of 502
-      if (error.response.status !== 502) {
+      if (error.response.status !== 502 && error.response.status !== 404) {
         return Promise.reject(error)
       }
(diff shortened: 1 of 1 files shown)


PVince81:
I assume that if the service is available but a different/unknown/inaccessible UUID is given, it would return a different code ? or would the error body make it possible to distinguish ?

if relevant, let's raise a ticket to address these cases separately

kulmann:
> I assume that if the service is available but a different/unknown/inaccessible UUID is given, it would return a different code ? or would the error body make it possible to distinguish ?
> 
> if relevant, let's raise a ticket to address these cases separately

Correct, if the service is not available, 404 or 502 happen and those are ignored (otherwise you would see an error in phoenix on every page reload, if you don't have the settings service or don't want to use it). Those two error codes must not be used on errors from within the settings service.

All other errors are passed through to Phoenix and will at the moment result in an error message in the ui, see: https://github.com/owncloud/phoenix/pull/3484#issuecomment-633183217
I think that is sufficient for now.

PR #494, opened 2020-06-25
Fixes #493

### LINKED ISSUE #493: uuid deprication warning
````
Deep requiring like `const uuidv4 = require('uuid/v4');` is deprecated as of uuid@7.x. Please require the top-level module when using the Node.js CommonJS module or use ECMAScript Modules when bundling for the browser. See https://github.com/uuidjs/uuid#deep-requires-now-deprecated for more information.

DeepDiver1975:


PVince81:
Aha, you found how to fix it ! Back then my quick fix did not work.

PR #504, opened 2020-07-02
Integrate owncloud-sign-url

DeepDiver1975:
This introduces url signing for owncloud - starting to be shipped as of 10.6(?)

```javascript
    // get an absolute url to a file - will work with webdav v1 as well
    const url = oc.files.getFileUrlV2('/file.txt')
    // sign the url - optional parameters are ttl and httpMethod
    const signedUrl = await oc.signUrl(url)
    // a simple fetch without additional headers can be done using the signed url
    const response = await fetch(signedUrl)
    expect(response.ok).toEqual(true)
    const txt = await response.text()

### DIFF
--- a/src/helperFunctions.js
+++ b/src/helperFunctions.js
@@ -6,6 +6,7 @@ const utf8 = require('utf8')
 const FileInfo = require('./fileInfo.js')
 const { v4: uuidv4 } = require('uuid')
 const HttpError = require('./httpError.js')
+const SignUrl = require('owncloud-sign-url')
 
 class helpers {
   constructor () {
@@ -34,6 +35,7 @@ class helpers {
     this._headers = {}
     this._versionNumber = null
     this._currentUser = null
+    this._signingKey = null
   }
 
   /**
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/helperFunctions.js [hunk 1]
```diff
@@ -131,6 +134,48 @@ class helpers {
...
+    const user = await this.getCurrentUser()
+    const signUrl = new SignUrl({
+      credential: user.id,
+      secretKey: key,
+      ttl: ttl,
+      algorithm: 'sha512'
+    })
+    try {
+      const signed = signUrl.generateSignedUrl(url, httpMethod)
+      return Promise.resolve(signed)
```
PVince81, line 156: you might not even need to resolve or reject, that's the magic of async.

if you return a value it will be considered a resolution

and if you throw an exception, it's like rejecting

best to double check though

DeepDiver1975:
> will you add the capability check in the SDK later ?

no need to add anything from my perspective.

Any consuming code base can already call oc.getCapabilities and check is the capability for signed url is there.

PR #538, opened 2020-08-20
owncloud.requests.ocs allows to add headers

DeepDiver1975:


src/helperFunctions.js
```diff
@@ -600,13 +600,15 @@ class helpers {
...
-      data: null
+      data: null,
+      headers: []
     }
     options = Object.assign({}, defaults, options)
     const action = options.action.includes('?') ? options.action + '&format=json' : options.action + '?format=json'
     const url = this.instance + this.OCS_BASEPATH_V2 + options.service + '/' + action
     const headers = this.buildHeaders()
     headers['OCS-APIREQUEST'] = 'true'
+    headers.concat(options.headers)
```
PVince81, line 611: headers = headers.concat(options.headers)

image

PVince81 (changes requested):

src/helperFunctions.js
PVince81: wait, I mixed it up with regular arrays

hash arrays don't have the `concat` option, did you test it ?

you could use `Object.assign()` instead like above

DeepDiver1975: yeah .... ci was already crying ..... bloody js

Issue #500, opened 2020-06-29
transform tests to pact.io


LukasHirt:
Would it make sense to run our unit tests also against oCIS backend? It'd be nice to prevent issues like the one we had now - https://github.com/owncloud/owncloud-sdk/pull/499 in future

LukasHirt:
Would it be too much effort for not much reward or a good idea? @individual-it

individual-it:
Not sure what the unit tests do here, but if there are unit tests then they should not depend on the backend :thinking:

phil-davis:
@LukasHirt please write some more about your thoughts.

LukasHirt:
> Not sure what the unit tests do here, but if there are unit tests then they should not depend on the backend

> please write some more about your thoughts.

With unit tests, we are testing here js wrappers around oC APIs. Mocking APIs would seem expensive here because we'd have to do it for almost every test. I think (but am not sure) that that's why there was introduced the running server in the unit tests and they are running against it. @individual-it @phil-davis

LukasHirt:
@individual-it @phil-davis Would there be some other way to do testing here? It'd be awesome if we could eliminate the running server from unit tests but am not sure how the SDK could be tested then without stumbling upon the issue from comment above. Ideas?

LukasHirt:
I've put the ticket into the next sprint input. Maybe we could schedule a short time to get this sorted out?

individual-it:
instead of running the whole server or  mocking it we could use contract testing, similar to what is done in https://github.com/owncloud/ocis-hello/pull/82

individual-it:
POC for contract tests in https://github.com/owncloud/owncloud-sdk/pull/543

swoichha:
###  Tests to be converted to pact.io
- [ ] -unauthenticated
    - [ ] - appTest.js
    - [ ] - capabilitiesTest.js
    - [ ] - filesTest.js
    - [ ] - groupTest.js
    - [ ] - sharingTest.js
    - [ ] - userTest.js

- [ ] -unauthorized
    - [ ] - appTest.js
    - [ ] - capabilitiesTest.js
    - [ ] - filesTest.js
    - [ ] - groupTest.js
    - [ ] - sharingTest.js
    - [ ] - userTest.js


 - [ ] - appTest.js
- [ ] - capabilitiesTest.js
- [ ] - filesTest.js
- [ ] - filesTrashTest.js
- [ ] - groupTest.js
- [ ] - loginTest.js
[...]

PR #543, opened 2020-08-31
tests with pact.io

individual-it:
transforming the tests to use contract based testing with pact.io as discussed in #500

this PR transforms 9 out of 256 tests and gets rid of codecov (obviously the coverage now is lower)


tests/requestsOcsTest.js
```diff
@@ -1,4 +1,4 @@
-describe('Main: Currently testing low level OCS', function () {
+fdescribe('Main: Currently testing low level OCS', function () {
```
LukasHirt, line 1: Is adding the "f" here a standard of pact or does it have some other reason?

LukasHirt (approved):
Just one comment to clarify something. Otherwise LGTM

tests/requestsOcsTest.js
individual-it: its there to only run the tests, that are already converted

Issue #544, opened 2020-08-31
only publish once


individual-it:
publishing should be shifted to an own pipeline that runs after and depends on the testing pipelines

individual-it:
CC @LukasHirt

LukasHirt:
Would it also make sense to switch to starlark here? It's quite simple so not sure if it'd have some real benefits but just to be consistent with other repos

PR #549, opened 2020-09-03
[Tests-Only] Convert drone.yml to drone.star

HariBhandari07:
### Description
Convert `.drone.yml` to `.drone.star`

### Related Issues
On top of https://github.com/owncloud/owncloud-sdk/pull/543
- Fixes https://github.com/owncloud/owncloud-sdk/issues/544


.drone.star
```diff
@@ -0,0 +1,235 @@
...
+		'master'
+	],
+
+	'yarnlint': True,
+
+	'build': True
+}
+
+def main(ctx):
+    return [ buildDocs() + buildSystem() + installServer() + configureServer() + owncloudLog() + waitForOwncloud() + prepareTestConfig() + test() + codecov() ]
```
dpakach, line 14: This should return the array of pipelines, not the steps

.drone.star
```diff
@@ -0,0 +1,235 @@
...
+			'/usr/local/bin/apachectl',
+			'-e',
+			'debug',
+			'-D',
+			'FOREGROUND'
+		]
+	}]
+
+
+def fullBuildAndPublish():
```
dpakach, line 220: This is not used anywhere. use this in main function.

.drone.star
```diff
@@ -0,0 +1,284 @@
...
+		'master'
+	],
+
+	'yarnlint': True,
+
+	'build': True
+}
+
+def main(ctx):
+    return [ fullBuildAndPublish() + testWithinSubFolder() ]
```
phil-davis, line 14: Might need to be:
```suggestion
    return [ fullBuildAndPublish(), testWithinSubFolder() ]
```

.drone.star
```diff
@@ -0,0 +1,173 @@
...
+            'branch': 'master'
+        },
+        'steps':
+            incrementVersion() +
+            buildDocs() +
+            buildSystem() +
+            prepareTestConfig() +
+            test() +
+            publishDocs() +
+            publishSystem()
```
individual-it, line 147: this

.drone.star
```diff
@@ -0,0 +1,173 @@
...
+    return [{
+        'name': 'test',
+        'image': 'owncloudci/chromium',
+        'pull': 'always',
+        'commands': [
+            'yarn test-drone'
+        ]
+    }]
+
+def publishDocs():
```
individual-it, line 79: this step will run twice, in fullBuildAndPublish and testWithinSubFolder, it should only run only once after both types of tests pass

.drone.star
```diff
@@ -0,0 +1,173 @@
...
+            }
+        },
+        'when': {
+            'event': [
+                'push'
+            ]
+        }
+    }]
+
+def publishSystem():
```
individual-it, line 102: same as publishDocs

individual-it (changes requested):

LukasHirt:
As we're switching to test with pact.io does it still make sense to run the tests in subfolder? Since we mock the responses there won't be much difference between those two anymore, will be?

individual-it:
@LukasHirt I think it makes sense as we are not testing the server but the consumer here. So if there is a chance that the SDK has a bug and reacting differently when sending a requests to `/sub/ocs/v1.php...` and not to `ocs/v1.php....` then it does make sense.
I believe we had such bugs in oc10. The backend would deliver correctly but the UI would not display the result correctly e.g. because its parsed the path wrong.
CC @PVince81 we talked about that in the chat recently

LukasHirt:
> I think it makes sense as we are not testing the server but the consumer here. So if there is a chance that the SDK has a bug and reacting differently when sending a requests to /sub/ocs/v1.php... and not to ocs/v1.php.... then it does make sense.

I see. Thanks for clearing out!

.drone.star
```diff
@@ -0,0 +1,195 @@
...
+            'event': [
+                'push'
+            ]
+        }
+    }]
+
+def fullBuildAndPublish():
+    return {
+        'kind': 'pipeline',
+        'name': 'Full build and publish',
```
individual-it, line 128: this pipeline does not publish anymore

.drone.star
```diff
@@ -0,0 +1,195 @@
...
+            },
+        },
+        'when': {
+            'event': [
+                'push'
+            ]
+        }
+    }]
+
+def fullBuildAndPublish():
```
individual-it, line 125: this pipeline does not publish anymore

individual-it (changes requested):

.drone.star
dpakach: That will only run after merge so we need to merge this Pr to see if that works.

individual-it (approved):

individual-it:
CC @LukasHirt I hope the publishing works

dpakach:
Publishing doesnt works currently.
Fixed in https://github.com/owncloud/owncloud-sdk/pull/561

PR #552, opened 2020-09-04
apps tests

individual-it:
part of #500
with this PR the status is:
``` 
Chrome Headless 85.0.4183.83 (Linux x86_64): Executed 17 of 256 (skipped 239) SUCCESS (8.284 secs / 1.134 secs)
TOTAL: 17 SUCCESS

### DIFF
--- a/tests/appsTest.js
+++ b/tests/appsTest.js
@@ -1,4 +1,4 @@
-describe('Main: Currently testing apps management,', function () {
+fdescribe('Main: Currently testing apps management,', function () {
 // CURRENT TIME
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/appsTest.js [hunk 1]
```diff
@@ -9,7 +9,7 @@ describe('Main: Currently testing apps management,', function () {
   var oc
 
   // TESTING CONFIGS
-  var testApp = 'someAppName' + timeRightNow
```
LukasHirt, line 12: I love how we don't have to create those random names any more!

LukasHirt (approved):
LGTM

PR #569, opened 2020-09-16
Convert filesTrash tests into pact.io

HariBhandari07:
part of #500 
blocked by https://github.com/pact-foundation/pact-js/issues/494


swoichha:
See:https://github.com/owncloud/owncloud-sdk/pull/591

PR #577, opened 2020-09-18
[Tests-Only] Convert sharing test to pact.io

jasson99:
## Related Issue:
- #500

### DIFF
--- a/tests/karma.js
+++ b/tests/karma.js
@@ -2014,7 +2014,7 @@ beforeAll(function (done) {
             '  <displayname_file_owner>admin</displayname_file_owner>\n' +
             '  <additional_info_owner/>\n' +
             '  <additional_info_file_owner/>\n' +
-            '  <path>/......123.txt</path>\n' +
+            '  <path>/.......txt</path>\n' +
             '  <item_type>file</item_type>\n' +
             '  <mimetype>text/plain</mimetype>\n' +
             '  <storage_id>home::admin</storage_id>\n' +
@@ -2022,7 +2022,7 @@ beforeAll(function (done) {
             '  <item_source>2147498361</item_source>\n' +
             '  <file_source>2147498361</file_source>\n' +
             '  <file_parent>3</file_parent>\n' +
-            '  <file_target>/......123.txt</file_target>\n' +
+            '  <file_target>/.......txt</file_target>\n' +
             '  <share_with>test123</share_with>\n' +
             '  <share_with_displayname>test123</share_with_displayname>\n' +
             '  <share_with_additional_info/>\n' +
@@ -2090,15 +2090,19 @@ beforeAll(function (done) {
         withRequest: {
           method: 'PUT',
           path: Pact.Matchers.regex({
-            matcher: '.*\\/remote\\.php\\/webdav\\/.*',
+            matcher: '.*\\/remote\\.php\\/webdav\\/' + config.testFile,
             generate: '/remote.php/webdav/' + config.testFile
           }),
           headers: validAuthHeaders,
-          body: '123456'
+          body: config.testContent
         },
         willRespondWith: {
           status: 200,
           headers: {
+            'Access-Control-Request-Method': Pact.Matchers.regex({
+              matcher: 'GET|POST|PUT|DELETE',
+              generate: 'GET'
+            }),
             'Access-Control-Allow-Origin': origin
           }
         }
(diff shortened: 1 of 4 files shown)


tests/karma.js
```diff
@@ -2151,6 +2155,1028 @@ beforeAll(function (done) {
...
+            '  <name>Öffentlicher Link</name>\n' +
+            '  <url>http://localhost/oc/s/H1wiMY2oDYmnWdC</url>\n' +
+            '  <attributes/>\n' +
+            ' </data>\n' +
+            '</ocs>'
+        }
+      }))
+    .then(() => {
+      provider.addInteraction({
+        uponReceiving: 'a request to update the permissions of a public link share',
```
individual-it, line 2945: please more details because this is an invalid request

tests/karma.js
```diff
@@ -2151,6 +2155,1028 @@ beforeAll(function (done) {
...
+        withRequest: {
+          method: 'PUT',
+          path: Pact.Matchers.regex({
+            matcher: '.*\\/ocs\\/v1\\.php\\/apps\\/files_sharing\\/api\\/v1\\/shares\\/9$',
```
individual-it, line 2949: make it more general e.g.
```suggestion
            matcher: '.*\\/ocs\\/v1\\.php\\/apps\\/files_sharing\\/api\\/v1\\/shares\\/\d+$',
```

tests/karma.js
```diff
@@ -2151,6 +2155,1028 @@ beforeAll(function (done) {
...
+            '  <itemsperpage></itemsperpage>\n' +
+            ' </meta>\n' +
+            ' <data/>\n' +
+            '</ocs>'
+        }
+      })
+    })
+    .then(() => {
+      provider.addInteraction({
+        uponReceiving: 'a request to update a public link share',
```
individual-it, line 2976: more details, this seems to be a very specific update

tests/karma.js
```diff
@@ -2151,6 +2155,1028 @@ beforeAll(function (done) {
...
+    )
+    .then(() => {
+      const shareids = [14, 18, 19]
+      for (const shareid of shareids) {
+        provider.addInteraction({
+          uponReceiving: 'a DELETE request for a share',
+          withRequest: {
+            method: 'DELETE',
+            path: Pact.Matchers.term({
+              matcher: '.*\\/ocs\\/v1\\.php\\/apps\\/files_sharing\\/api\\/v1\\/shares\\/' + shareid + '$',
```
individual-it, line 2290: does the id really matter? the response is always the same, so maybe we can get away with just a single interaction that matches all ids

individual-it (changes requested):
you can use existing variables in many places e.g. `validAuthHeaders` also I think some interactions can be made more general

tests/karma.js
```diff
@@ -2151,6 +2155,1028 @@ beforeAll(function (done) {
...
+    )
+    .then(() => {
+      const files = ['test123.txt', '%E6%96%87%E4%BB%B6123.txt', 'test%20space%20and%20%2B%20and%20%23123.txt', 'newFileCreated123']
+      for (const file of files) {
+        provider.addInteraction({
+          uponReceiving: 'a DELETE request for a file',
+          withRequest: {
+            method: 'DELETE',
+            path: Pact.Matchers.term({
+              matcher: '.*\\/remote\\.php\\/webdav\\/' + file + '$',
```
kiranparajuli589, line 2378:
```suggestion
              matcher: '.*\\/remote\\.php\\/webdav\\/*',
```
Is it ok to use matcher like this so that files with any filename can be deleted!

tests/karma.js
```diff
@@ -2151,6 +2155,1028 @@ beforeAll(function (done) {
...
+            '<ocs>\n' +
+            ' <meta>\n' +
+            '  <status>failure</status>\n' +
+            '  <statuscode>404</statuscode>\n' +
+            '  <message>Wrong share ID, share doesn\'t exist</message>\n' +
+            '  <totalitems></totalitems>\n' +
+            '  <itemsperpage></itemsperpage>\n' +
+            ' </meta>\n' +
+            ' <data/>\n' +
+            '</ocs>'
```
kiranparajuli589, line 2792: useful if saved is some variable

tests/karma.js
jasson99: this gives error about multiple interactions. this can be fixed once the refactoring work comes in progress

tests/karma.js
jasson99: fixed

tests/karma.js
jasson99: this gives multiple interaction error, which can be fixed once the refactoring work comes into progress

tests/karma.js
jasson99: fixed

tests/karma.js
jasson99: fixed

tests/karma.js
jasson99: may be we can fix that in the refactoring

individual-it (approved):

PR #583, opened 2020-09-25
refactor pact.io tests to attach interactions to the tests

individual-it:
interactions should be attached to the specific tests, and should not all being loaded for all tests. 
I understood this after stumbling of an issue that I could not solve in the old system (see https://github.com/pact-foundation/pact-js/issues/511) and then talking to people of pact.io 

This PR is the first step into this direction.
1. put all existing interaction into a helper
2. add specific interactions to public link tests

tasks for other PRs:
- move interactions that are not needed in every test to the specific test(s)
- use `verify` and not `removeInteractions` this is not possible now because not all interactions are actually used, so `verify` fails

should be merged after  #577


tests/sharingTest.js
```diff
@@ -732,6 +745,8 @@ fdescribe('Main: Currently testing file/folder sharing,', function () {
           testUser,
           { expirationDate: expirationDate }
         ).then(share => {
+          console.log(share.getExpiration())
+          console.log(typeof share.getExpiration())
```
phil-davis, line 749: Were these intended to be added?

tests/publicLinkTest.js
```diff
@@ -97,7 +106,88 @@ describe('oc.publicFiles', function () {
...
+          promises.push(provider.addInteraction({
+            uponReceiving: 'create a public link share',
+            withRequest: {
+              method: 'POST',
+              path: Pact.Matchers.regex({
+                matcher: '.*\\/ocs\\/v(1|2)\\.php\\/apps\\/files_sharing\\/api\\/v1\\/shares',
+                generate: '/ocs/v1.php/apps/files_sharing/api/v1/shares'
+              }),
+              headers: validAuthHeaders
+              // body: 'shareType=3&path=%2F' + config.testFolder
```
phil-davis, line 121: I guess that you do not [want|need] to be specific about the body when matching here.
Do you want to leave this as a comment?

phil-davis (approved):

PR #585, opened 2020-09-28
[Tests-Only] Refactor PactHelper

jasson99:
## Description:
 - This PR refactors the pactHelper.js to make the code DRY and maintainable. The common lines of code shared by different interactions such as response body, responseMeta etc are adjusted accordingly.

## Related Issue:
- Fixes https://github.com/owncloud/owncloud-sdk/issues/587

### DIFF
--- a/tests/pactHelper.js
+++ b/tests/pactHelper.js
@@ -10,11 +10,40 @@ const validAuthHeaders = {
   authorization: 'Basic ' + validUserPasswordHash,
   Origin: origin
 }
-const ocsSuccessMeta = ' <meta>\n' +
-  '  <status>ok</status>\n' +
-  '  <statuscode>100</statuscode>\n' +
-  '  <message/>\n' +
-  ' </meta>\n'
+
+const ocsMeta = function (status, statusCode, Message = null) {
+  if (Message == null) {
+    return ' <meta>\n' +
+      '  <status>' + status + '</status>\n' +
+      '  <statuscode>' + statusCode + '</statuscode>\n' +
+      '  <message/>\n' +
+      ' </meta>\n'
+  }
+  return ' <meta>\n' +
+    '  <status>' + status + '</status>\n' +
+    '  <statuscode>' + statusCode + '</statuscode>\n' +
+    '  <message>' + Message + '</message>\n' +
+    ' </meta>\n'
+}
+
+const shareResponseOcsData = function (shareType, id, permissions, fileTarget) {
+  const data = '  <id>' + id + '</id>\n' +
+    '  <share_type>' + shareType + '</share_type>\n' +
+    '  <uid_owner>admin</uid_owner>\n' +
+    '  <displayname_owner>admin</displayname_owner>\n' +
+    '  <permissions>' + permissions + '</permissions>\n' +
+    '  <uid_file_owner>admin</uid_file_owner>\n' +
+    '  <displayname_file_owner>admin</displayname_file_owner>\n' +
+    '  <path>' + fileTarget + '</path>\n' +
+    '  <file_target>' + fileTarget + '</file_target>\n'
+
+  if (shareType === 3) {
+    return data +
+      '  <url>' + config.owncloudURL + '/yrkoLeS33y1aTya</url>\n'
+  }
+  return data
+}
+
 const xmlResponseHeaders = {
   'Content-Type': 'text/xml; charset=utf-8',
   'Access-Control-Allow-Origin': origin
@@ -24,6 +53,7 @@ const invalidAuthHeader = Pact.Matchers.term({
   matcher: '^(?!Basic ' + validUserPasswordHash + ').*$', // match anything except a valid auth
   generate: 'Basic bm9uRXhpc3RpbmdVc2VycnByeXJxOHg2OmNvbmZpZy5wYXNzd29yZHJwcnlycTh4Ng=='
 })
+
 const unauthorizedXmlResponseBody = '<?xml version="1.0"?>\n' +
   '<ocs>\n' +
   ' <meta>\n' +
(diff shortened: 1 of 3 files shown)


tests/pactHelper.js
```diff
@@ -10,11 +10,40 @@ const validAuthHeaders = {
...
+    '  <displayname_owner>admin</displayname_owner>\n' +
+    '  <permissions>' + permissions + '</permissions>\n' +
+    '  <uid_file_owner>admin</uid_file_owner>\n' +
+    '  <displayname_file_owner>admin</displayname_file_owner>\n' +
+    '  <path>' + fileTarget + '</path>\n' +
+    '  <file_target>' + fileTarget + '</file_target>\n'
+
+  if (shareType === 3) {
+    return data +
+      '  <url>http://localhost/oc/s/yrkoLeS33y1aTya</url>\n'
```
individual-it, line 42: this is not true for my setup nor for the CI setup. Is that data even needed?

tests/pactHelper.js
```diff
@@ -2186,7 +2068,7 @@ function setGeneralInteractions (provider) {
         }),
         query: 'path=' + path[i],
         headers: {
-          authorization: 'Basic YWRtaW46YWRtaW4=',
+          authorization: 'Basic ' + validUserPasswordHash,
```
individual-it, line 2071: why not using `validAuthHeaders`?

individual-it (changes requested):

tests/pactHelper.js
jasson99: fixed

tests/pactHelper.js
jasson99: Yes, the link is expected to be string. I have used the config.owncloudUrl now

tests/pactHelper.js
```diff
@@ -164,13 +194,7 @@ function setGeneralInteractions (provider) {
...
         matcher: '<\\?xml version="1\\.0"\\?>\\s' +
           '<ocs>\\s' +
-          ' <meta>\\s' +
-          '  <status>ok<\\/status>\\s' +
-          '  <statuscode>100<\\/statuscode>\\s' +
-          '  <message>OK<\\/message>\\s' +
-          '  <totalitems><\\/totalitems>\\s' +
-          '  <itemsperpage><\\/itemsperpage>\\s' +
-          ' <\\/meta>\\s' +
+          ocsMeta('ok', 100, 'OK') +
```
kiranparajuli589, line 197: you've to also add pagination tags in `ocsMeata`
```js
'  <totalitems><\\/totalitems>\\s' +
'  <itemsperpage><\\/itemsperpage>\\s' +
```

kiranparajuli589, line 197: you've to also add pagination tags in `ocsMeata` or better use separate constant for paginated `xml` response
```js
'  <totalitems><\\/totalitems>\\s' +
'  <itemsperpage><\\/itemsperpage>\\s' +
```

jasson99: It is not required in most of the responses. If it is required later in any, we can just add it there in the specific response or can create a new var for that then

tests/pactHelper.js
```diff
@@ -10,11 +10,40 @@ const validAuthHeaders = {
...
+
+  if (shareType === 3) {
+    return data +
+      '  <url>' + config.owncloudURL + '/yrkoLeS33y1aTya</url>\n'
+  }
+  return data
+}
+
 const xmlResponseHeaders = {
   'Content-Type': 'text/xml; charset=utf-8',
```
kiranparajuli589, line 48: use vars for different content types
```
'Content-Type': textWithHtml
```

tests/pactHelper.js
```diff
@@ -260,7 +278,7 @@ function setGeneralInteractions (provider) {
       headers: xmlResponseHeaders,
       body: '<?xml version="1.0"?>\n' +
         '<ocs>\n' +
-        ocsSuccessMeta +
+        ocsMeta('ok', '100') +
         '  <data/>\n' +
         '</ocs>\n'
     }
```
kiranparajuli589, line 284: IMO covers more duplication if this can be managed like:
```
body: ocsResponseXml('ok', '100')
```

kiranparajuli589:
There is lots of 
```
'Access-Control-Allow-Methods': 'GET,OPTIONS,POST,PUT,DELETE,MKCOL,PROPFIND,PATCH,PROPPATCH,REPORT'
```
inside pactHelper. this can also be used from some `variable`

tests/pactHelper.js
```diff
@@ -10,11 +10,40 @@ const validAuthHeaders = {
...
+    '  <displayname_owner>admin</displayname_owner>\n' +
+    '  <permissions>' + permissions + '</permissions>\n' +
+    '  <uid_file_owner>admin</uid_file_owner>\n' +
+    '  <displayname_file_owner>admin</displayname_file_owner>\n' +
+    '  <path>' + fileTarget + '</path>\n' +
+    '  <file_target>' + fileTarget + '</file_target>\n'
+
+  if (shareType === 3) {
+    return data +
+      '  <url>' + config.owncloudURL + '/yrkoLeS33y1aTya</url>\n'
```
individual-it, line 42: will that token always be the same?

individual-it (approved):
LGTM as first step

tests/pactHelper.js
jasson99: that might create confusions later. So i think it is better to do as it is done now, since the code inside the meta part is similar almost everywhere with just different value of status and status code. the other elements such as data and  element are not easy to manage in the same function or variable

tests/pactHelper.js
jasson99: No, it is usually different for different response. But that does not really matter, IMO. All that is checked is if the url is a string.

PR #591, opened 2020-10-01
[Tests-Only] Convert fileTrashTest.js to pact.io

swoichha:
Convert `tests/fileTrashTest.js` to pact.io

## Related Issue:
part of #500

### DIFF
--- a/tests/config/config.drone.json
+++ b/tests/config/config.drone.json
@@ -36,5 +36,5 @@
       "JbdgMGkt3Cq6B0u"
     ],
     "testContent": "123456",
-    "expirationDate": "2022-10-01"
+  "expirationDate": "2022-10-01"
 }
--- a/tests/fileTrashTest.js
+++ b/tests/fileTrashTest.js
@@ -1,24 +1,123 @@
-describe('oc.fileTrash', function () {
+fdescribe('oc.fileTrash', function () {
   const OwnCloud = require('../src/owncloud')
   const config = require('./config/config.json')
 
   // LIBRARY INSTANCE
   let oc
-  let trashEnabled = false
+  const trashEnabled = true
   const userId = config.username
 
   // PACT setup
   const Pact = require('@pact-foundation/pact-web')
   const provider = new Pact.PactWeb()
-  const { setGeneralInteractions } = require('./pactHelper.js')
+  const { origin, validAuthHeaders, accessControlAllowHeaders, accessControlAllowMethods, setGeneralInteractions } = require('./pactHelper.js')
+  const deletedFolderId = '2147596415'
+  const deletedFileId = '2147596419'
 
-  beforeAll(function (done) {
-    Promise.all(setGeneralInteractions(provider)).then(done, done.fail)
+  const trashbinPath = Pact.Matchers.term({
+    matcher: '.*\\/remote\\.php\\/dav\\/trash-bin\\/admin\\/\\/$',
+    generate: '/remote.php/dav/trash-bin/admin//'
   })
 
-  afterAll(function (done) {
-    provider.removeInteractions().then(done, done.fail)
+  const trashbinFolderPath = Pact.Matchers.term({
+    matcher: '.*\\/remote\\.php\\/dav\\/trash-bin\\/admin\\/' + deletedFolderId,
+    generate: '/remote.php/dav/trash-bin/admin/' + deletedFolderId
   })
+
+  const responseHeader = function (contentType) {
+    return (
+      {
+        'Access-Control-Allow-Origin': origin,
+        'Access-Control-Allow-Headers': accessControlAllowHeaders,
+        'Access-Control-Allow-Methods': accessControlAllowMethods,
+        'Content-Type': contentType
+      }
+    )
+  }
+
+  const emptyTrashbinXmlRequestBody = '<?xml version="1.0"?>\n' +
+    '<d:propfind  xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">\n' +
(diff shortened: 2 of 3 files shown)


tests/fileTrashTest.js
```diff
@@ -177,18 +463,135 @@ describe('oc.fileTrash', function () {
...
+            headers: {
+              'Content-Type': 'text/html; charset=utf-8',
+              'Access-Control-Allow-Origin': origin,
+              'Access-Control-Allow-Headers': accessControlAllowHeaders,
+              'Access-Control-Allow-Method': accessControlAllowMethods
+            }
+          }
+        }))
+        promises.push(provider.addInteraction({
+          uponReceiving: 'PROPFIND to an empty trashbin after restoring',
```
individual-it, line 494: looks same like `'PROPFIND to an empty trashbin',` can we use one interaction for both?

tests/fileTrashTest.js
```diff
@@ -207,61 +610,231 @@ describe('oc.fileTrash', function () {
...
+            headers: {
+              'Content-Type': 'text/html; charset=utf-8',
+              'Access-Control-Allow-Origin': origin,
+              'Access-Control-Allow-Headers': accessControlAllowHeaders,
+              'Access-Control-Allow-Method': accessControlAllowMethods
+            }
+          }
+        }))
+        promises.push(provider.addInteraction({
+          uponReceiving: 'PROPFIND trash items after restoring deleting file',
```
individual-it, line 736: looks same like `'PROPFIND to an empty trashbin',`

tests/config/config.drone.json
```diff
@@ -36,5 +36,7 @@
       "JbdgMGkt3Cq6B0u"
     ],
     "testContent": "123456",
-    "expirationDate": "2022-10-01"
+  "expirationDate": "2022-10-01",
+  "deletedFolderId": "2147596415",
+  "deletedFileId": "2147596419"
```
individual-it, line 41: if those values are only used in `tests/fileTrashTest.js` then lets put them there

individual-it (changes requested):
some interactions look like they are dublicated also you could do some  path more dynamic `.../admin/..`

tests/fileTrashTest.js
swoichha: Yes, we can use one interaction for both. Need to do some refactoring for that which is WIP.

tests/fileTrashTest.js
```diff
@@ -177,18 +463,135 @@ describe('oc.fileTrash', function () {
...
+            }),
+            headers: {
+              Destination: config.owncloudURL + 'remote.php/dav/files/admin/' + config.testFolder + '%20(restored%20to%20a%20different%20location)'
+            }
+          },
+          willRespondWith: {
+            status: 201,
+            headers: {
+              'Content-Type': 'text/html; charset=utf-8',
+              'Access-Control-Allow-Origin': origin,
```
jasson99, line 487: we have variable for this

jasson99 (changes requested):

tests/fileTrashTest.js
```diff
@@ -60,106 +124,291 @@ describe('oc.fileTrash', function () {
...
+              '</d:propfind>'
+          },
+          willRespondWith: {
+            status: 207,
+            headers: {
+              'Access-Control-Allow-Origin': origin,
+              'Access-Control-Allow-Headers': accessControlAllowHeaders,
+              'Access-Control-Allow-Methods': accessControlAllowMethods,
+              'Content-Type': 'text/html; charset=utf-8'
+            },
```
kiranparajuli589, line 383: I think this is also used in multiple interactions! You can give it a name too.

tests/fileTrashTest.js
```diff
@@ -176,20 +425,137 @@ describe('oc.fileTrash', function () {
...
+        promises.push(provider.addInteraction({
+          uponReceiving: 'MOVE folder from trashbin to fileslist to a different location',
+          withRequest: {
+            method: 'MOVE',
+            path: Pact.Matchers.term({
+              matcher: '.*\\/remote\\.php\\/dav\\/trash-bin\\/admin\\/' + config.deletedFolderId,
+              generate: '/remote.php/dav/trash-bin/admin/' + config.deletedFolderId
+            }),
+            headers: {
+              Destination: config.owncloudURL + 'remote.php/dav/files/admin/' + config.testFolder + '%20(restored%20to%20a%20different%20location)'
```
kiranparajuli589, line 443: you can use a matcher to compare the destination as in `path`

individual-it (approved):

tests/fileTrashTest.js
```diff
@@ -60,106 +143,228 @@ describe('oc.fileTrash', function () {
...
-      testFolder = 'testFolder' + suffix
-      testFile = testFolder + '/file.txt'
-      oc.files.createFolder(testFolder).then(() => {
-        oc.files.putFileContents(testFile, '*').then(() => {
-          oc.files.delete(testFolder).then(() => {
-            done()
-          })
-        })
-      })
+  fdescribe('and when empty', function () {
```
jasson99, line 167: You could make it more clear. something like `when the trashbin is empty`

swoichha: Yes, but originally it was `and when empty`. So I'm not sure if we should change it or not https://github.com/owncloud/owncloud-sdk/blob/26bc93b166d001835a4995249af0bb207729818c/tests/fileTrashTest.js#L77

jasson99 (commented):
if you want all the tests in this file to run, then you can simply let the tests be as `it` instead of `fit`.

PR #593, opened 2020-10-07
Refactore unauthorized files test to include interactions inside test file

kiranparajuli589:
## Description﻿
With this PR:
- Included required interactions for `unauthorized/appTest.js` in the same test file

## Related Issue
- Part of #500


jasson99 (commented):
I think you can keep those `it` as it is instead of making `fit` if all the tests are to be run.

individual-it (approved):

PR #594, opened 2020-10-07
Refactored unauthorized capabilities test

kiranparajuli589:
## Description
With this PR:
- Required interaction for `unauthorized/capabilitiesTest.js`﻿ are moved in the same test file

## Related Issue
- #500


individual-it:
tests fail :-(
image

kiranparajuli589:
> tests fail :-(

Yes, its because the moved interaction is also required for test in `loginTest`. So it is better not to move the interaction from the `pactHelper.js`

PR #603, opened 2020-10-14
[Tests-Only] Interactions for apps test added into apps test file

jasson99:
## Related Issue:
- #500


tests/appsTest.js
```diff
@@ -15,10 +15,133 @@ fdescribe('Main: Currently testing apps management,', function () {
...
+          headers: validAuthHeaders
+        },
+        willRespondWith: {
+          status: 200,
+          headers: xmlResponseHeaders,
+          body: '<?xml version="1.0"?>\n' +
+            '<ocs>\n' +
+            ocsMeta('ok', '100') +
+            data +
+            '</ocs>'
```
swoichha, line 114: We can define this into a constant to make the code DRY as it is repeated elsewhere as well.

tests/appsTest.js
```diff
@@ -131,6 +279,42 @@ fdescribe('Main: Currently testing apps management,', function () {
...
```
swoichha, line 312: same

swoichha (changes requested):

tests/appsTest.js
```diff
@@ -15,10 +15,133 @@ fdescribe('Main: Currently testing apps management,', function () {
...
+        for (const [key, value] of Object.entries(attributes[attribute].value)) {
+          data = data +
+            '  <element>\n' +
+            '   <key>' + utf8.encode(key) + '</key>\n' +
+            '   <app>someAppName</app>\n' +
+            '   <value>' + utf8.encode(value) + '</value>\n' +
+            '  </element>\n'
+        }
+        data = data + ' </data>'
+      } else if (attributes[attribute].attrExists === true) {
```
kiranparajuli589, line 87:
```suggestion
      } else if (attributes[attribute].attrExists) {
```

tests/appsTest.js
```diff
@@ -15,10 +15,133 @@ fdescribe('Main: Currently testing apps management,', function () {
...
+        attrExists: true,
+        value: ''
+      }
+    }
+    for (const attribute in attributes) {
+      // default no data
+      let data = ' <data/>\n'
+
+      // no attributes specified, return all attributes
+      if (attribute === '') {
```
kiranparajuli589, line 76:
```suggestion
      if (!attribute) {
```

tests/appsTest.js
jasson99: fixed

tests/appsTest.js
jasson99: fixed

tests/appsTest.js
jasson99: fixed

tests/appsTest.js
```diff
@@ -9,16 +7,138 @@ fdescribe('Main: Currently testing apps management,', function () {
...
-  var testApp = 'someAppName'
-  var nonExistentApp = 'nonExistentApp' + timeRightNow
+  var nonExistentApp = 'nonExistentApp123'
 
   // PACT setup
   const Pact = require('@pact-foundation/pact-web')
   const provider = new Pact.PactWeb()
-  const { setGeneralInteractions } = require('./pactHelper.js')
+  const { setGeneralInteractions, ocsMeta } = require('./pactHelper.js')
+  const { validAuthHeaders, xmlResponseHeaders } = require('./pactHelper.js')
```
swoichha, line 16: You can do something like this
```suggestion
   const {
    setGeneralInteractions,
    ocsMeta,
    validAuthHeaders,
    xmlResponseHeaders
  } = require('./pactHelper.js')
```

HariBhandari07 (commented):
I think it's nice to replace all `var`s with `let`s and `const`s as it is the modern way of js and it also makes the variable declaration in the PR more consistent.

tests/appsTest.js
```diff
@@ -9,16 +7,137 @@ fdescribe('Main: Currently testing apps management,', function () {
...
   // TESTING CONFIGS
-  var testApp = 'someAppName'
-  var nonExistentApp = 'nonExistentApp' + timeRightNow
+  var nonExistentApp = 'nonExistentApp123'
 
   // PACT setup
   const Pact = require('@pact-foundation/pact-web')
   const provider = new Pact.PactWeb()
-  const { setGeneralInteractions } = require('./pactHelper.js')
+  const { validAuthHeaders, xmlResponseHeaders, setGeneralInteractions, ocsMeta } = require('./pactHelper.js')
```
jasson99, line 15: .

tests/appsTest.js
jasson99: I prefer as it is now. I don't want to add extra lines to import the same thing.

swoichha (commented):
LGTM :+1:

swoichha (approved):

individual-it (approved):

PR #604, opened 2020-10-14
[Tests-Only] Interactions for share Recipient test added into shareRecipient file

jasson99:
## Related Issue:
- #500

### DIFF
--- a/tests/shareRecipientTest.js
+++ b/tests/shareRecipientTest.js
@@ -8,7 +8,7 @@ fdescribe('Main: Currently testing share recipient,', function () {
   // PACT setup
   const Pact = require('@pact-foundation/pact-web')
   const provider = new Pact.PactWeb()
-  const { setGeneralInteractions } = require('./pactHelper.js')
+  const { setGeneralInteractions, validAuthHeaders, origin } = require('./pactHelper.js')
 
   beforeAll(function (done) {
     Promise.all(setGeneralInteractions(provider)).then(done, done.fail)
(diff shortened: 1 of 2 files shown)


swoichha (commented):
Why is there only one interaction added in `shareRecipientTest` file? Is it WIP or we won't be needing any interaction for remaining three tests?

tests/shareRecipientTest.js
```diff
@@ -93,7 +69,33 @@ fdescribe('Main: Currently testing share recipient,', function () {
...
+        headers: {
+          'Content-Type': 'application/json; charset=utf-8',
+          'Access-Control-Allow-Origin': origin
+        },
+        body: '{"ocs"' +
+          ':{"meta":{"status":"ok","statuscode":200,' +
+          '"message":"OK","totalitems":"","itemsperpage":""},' +
+          '"data":{"exact":{"users":[],"groups":[],"remotes":[]},' +
+          '"users":[{"label":"test123","value":{"shareType":0,"shareWith":"test123"}}],' +
+          '"groups":[{"label":"testGroup","value":{"shareType":1,"shareWith":"testGroup"}}],' +
```
individual-it, line 95: can we make the username and groupname dynamic

tests/shareRecipientTest.js
```diff
@@ -19,7 +19,6 @@ fdescribe('Main: Currently testing share recipient,', function () {
   })
 
   // TESTING CONFIGS
-  const testUserPassword = config.testUserPassword
   const testUser = config.testUser
   const testGroup = config.testGroup
```
individual-it, line 23: why redefining the variables

individual-it (changes requested):

tests/shareRecipientTest.js
jasson99: fixed

tests/shareRecipientTest.js
jasson99: fixed

individual-it (approved):

PR #605, opened 2020-10-14
[Tests-Only] Interactions for group test added inside loginTest file

swoichha:
This PR adds the interactions related to group tests into the `tests/loginTest.js` file.

## Related Issue:
- #500


individual-it (approved):

swoichha:
Since the moved interaction is also required for tests in `unauthorized/capabilitiesTest`. So, it is better not to move the interaction from the `pactHelper.js`

PR #612, opened 2020-10-20
Refactor unauthorized user tests file

jasson99:
## Description:
In this PR, the interactions for the unauthorized user tests which were previously added in pact Helper file is maintained in the test file itself.

## Related Issue:
- #500


individual-it (approved):
I would be interested to know if the tests actually also run faster now, they should, because we add less interactions before every test

PR #614, opened 2020-10-23
[Tests-Only] Refactor file test

HariBhandari07:
### Description
This PR transforms all tests in `filesTest.js` to pact.io.

### Related Issue
Related to https://github.com/owncloud/owncloud-sdk/issues/500

### DIFF
--- a/tests/filesTest.js
+++ b/tests/filesTest.js
@@ -1,6 +1,4 @@
 fdescribe('Main: Currently testing files management,', function () {
-  // CURRENT TIME
-  const timeRightNow = Math.random().toString(36).substr(2, 9)
   const FileInfo = require('../src/fileInfo')
   const OwnCloud = require('../src/owncloud')
   const config = require('./config/config.json')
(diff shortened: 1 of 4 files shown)


swoichha (commented):
LGTM

tests/filesTest.js
```diff
@@ -202,345 +337,362 @@ fdescribe('Main: Currently testing files management,', function () {
...
-      expect(error).toBe(null)
-      done()
+      oc.files.mkdir(newFolder).then(status => {
+        expect(status).toBe(true)
+        return oc.files.list(newFolder, 0)
+      }).then(folder => {
+        folder = folder[0]
+        expect(folder.isDir()).toBe(true)
+        expect(folder.getName()).toEqual('new folder')
+        return oc.files.delete(newFolder)
```
individual-it, line 369: no need to delete the folder after the test, this test is about mkdir and there is no real server running where the folder is created. so this "delete" test should be moved to a separate test, or delete it completely if there are other explicit delete tests (can be done in a separate PR)

tests/filesTest.js
```diff
@@ -202,345 +337,362 @@ fdescribe('Main: Currently testing files management,', function () {
...
-    }).catch(error => {
-      expect(error.message).toBe('File with name ' + config.nonExistentDir.slice(1) + ' could not be located')
-      done()
+      oc.files.mkdir(newFolder).then(status => {
+        expect(status).toBe(true)
+        return oc.files.list(newFolder, 0)
+      }).then(folder => {
+        folder = folder[0]
+        expect(folder.isDir()).toBe(true)
+        expect(folder.getName()).toEqual('subdir')
```
individual-it, line 419: no need to create those folders, they are anyway not created for real

tests/filesTest.js
```diff
@@ -202,345 +337,362 @@ fdescribe('Main: Currently testing files management,', function () {
...
+      }).then(status2 => {
+        expect(status2).toEqual(true)
+        return oc.files.list(newFolder, 0)
+      }).then(folder2 => {
+        fail(folder2)
+        done()
+      }).catch(error => {
+        expect(error.message).toBe('File with name ' + newFolder + ' could not be located')
+        done()
+      })
```
individual-it, line 430: no need to list those folders afterwards

individual-it (changes requested):
looks generally good but:
1. write more comments about what you have done in the PR description (comments are love-letters to your future-self)
2. there are still setup and teardown jobs done (e.g. create a folder to test deletion and visa-verse) - we don't need them because we only run the tests on a mock server. make explicit mkdir and explicit delete tests

individual-it (approved):

PR #626, opened 2020-12-02
[Tests-Only] Run sign url test

HariBhandari07:
### Description
This PR
1. matches the hashed key in `signed url generator` and `signed url verifier` . Because of https://github.com/owncloud/owncloud-sdk/pull/616, hashed key doesn't match in `generateSignedUrl` and `verifySignedUrl` function
2. runs `owncloud-sign-url/signUrlTest.js`

### Related Issue
- On top of https://github.com/owncloud/owncloud-sdk/pull/616
- Part of https://github.com/owncloud/owncloud-sdk/issues/500

### DIFF
--- a/tests/owncloud-sign-url/signUrlTest.js
+++ b/tests/owncloud-sign-url/signUrlTest.js
@@ -6,7 +6,7 @@ const HTTP_GET_METHOD = 'get'
 const HTTP_POST_METHOD = 'post'
 const TEST_PORT = 33001
 
-describe('SignUrl tests', () => {
+fdescribe('SignUrl tests', () => {
   let signUrl
 
   beforeAll(async () => {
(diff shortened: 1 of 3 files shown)


tests/owncloud-sign-url/signUrlTest.js
```diff
@@ -6,7 +6,7 @@ const HTTP_GET_METHOD = 'get'
 const HTTP_POST_METHOD = 'post'
 const TEST_PORT = 33001
 
-describe('SignUrl tests', () => {
+fdescribe('SignUrl tests', () => {
```
DeepDiver1975, line 9: confusing - title states 'enable tests' but the test is disabled?

DeepDiver1975 (changes requested):

tests/owncloud-sign-url/signUrlTest.js
individual-it: the `f` enables the test. As we are in the process of changing tests over to pact.io tests that should run have an `f` infront of `describe` when all tests are converted all the `f`s will be deleted

individual-it (approved):

LukasHirt (approved):

DeepDiver1975 (approved):

PR #637, opened 2020-12-28
move tests to jest & pactV3 (capabilitiesTest.js)

individual-it:
move `capabiliesTest.js` to pactV3 so that we can use `XmlBuilder` to create the xml responses see https://github.com/pact-foundation/pact-js#using-pact-with-xml

to use pactV3 we need to move the tests to jest first

main things that need to be refactored:
- files that are converted need to be added in `jest.config.js` into `testMatch` (when all files are converted we can change that string to a regex that matches all files)
- interactions are now added with a chainable object `provider.uponReceiving('...').withRequest({...}).      .willRespondWith({...})`
- tests are placed inside a callback of `provider.executeTest()`
- CORS settings are not needed anymore (preflight, Access-Control headers)
- regular expression matcher takes now two arguments and not an object
- regular expression matcher can take a real [regular expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions) or a string, by using the regular expression we can get rid of the ugly `\\/`

part of #645 & #646


dpakach (commented):
LGTM :+1:

dpakach (approved):

PR #638, opened 2020-12-30
[Tests-Only]Refactor the capabilities test file to use provider verify

jasson99:
## Description:
This PR refactors the capabilities test file such that only the required interactions are imported. Also, verification of the requests and the interactions is done.

## Related Issue:
- Part of #643

### DIFF
--- a/tests/capabilitiesTest.js
+++ b/tests/capabilitiesTest.js
@@ -1,19 +1,41 @@
 describe('Main: Currently testing getConfig, getVersion and getCapabilities', function () {
-  const OwnCloud = require('../src/owncloud')
-  const config = require('./config/config.json')
-
-  // LIBRARY INSTANCE
   let oc
-
   // PACT setup
   const Pact = require('@pact-foundation/pact-web')
   const provider = new Pact.PactWeb()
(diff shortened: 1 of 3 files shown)


jasson99:
The interactions for which functions are created and exported are not removed yet, because they are still used in other test files. Once we refactor all or almost all the test files, we can get rid of the redundancy.

HariBhandari07:
LGTM

tests/capabilitiesTest.js
```diff
@@ -8,12 +8,52 @@ describe('Main: Currently testing getConfig, getVersion and getCapabilities', fu
...
+      auth: {
+        basic: {
+          username: config.username,
+          password: config.password
+        }
+      }
+    })
+    await oc.login()
+    return oc
+  }
```
individual-it, line 32: this is repeated in every file, can we move it to the helper?

individual-it (changes requested):

tests/capabilitiesTest.js
jasson99: fixed

individual-it (approved):

Issue #643, opened 2021-01-05
verify consumer side tests


individual-it:
currently we are creating a lot of interactions that are not used in the tests
running `verify()` will show the issue
we need to reduce the  created interactions to match what is really sent

see POC: #640

HariBhandari07:
###  Tests to be verified
- [x] -unauthenticated (nothing need to be done, in this tests no request is send to the server)
    - [x] - appTest.js
    - [x] - capabilitiesTest.js
    - [x] - filesTest.js
    - [x] - groupTest.js
    - [x] - sharingTest.js
    - [x] - userTest.js

- [x] -unauthorized
    - [x] - appTest.js 
    - [x] - capabilitiesTest.js  
    - [x] - filesTest.js  
    - [x] - groupTest.js 
    - [x] - sharingTest.js 
    - [x] - userTest.js 

- [x] - appTest.js 
- [x] - capabilitiesTest.js 
- [x] - filesTest.js 
- [x] - filesTrashTest.js
- [x] - groupTest.js 
- [x] - loginTest.js 
- [x] - publickLinkTest.js 
- [x] - requestsOcsTest.js 
- [x] - shareRecipientTest.js 
- [x] - sharingTest.js 
- [x] - sharingWithAttributesTest.js 
- [x] - signedUrlTest.js 
- [x] - signedUrlIntegrationTest.js 
- [x] - userTest.js 
- [x] - xmlParserTest.js  (nothing need to be done, in this tests no request is send to the server)

dpakach:
done :+1:

PR #654, opened 2021-01-07
[Tests-Only] Verify unauthorized tests

HariBhandari07:
## Description:
This PR refactors unauthorized tests such that only the required interactions are imported. Also, verification of the requests and the interactions is done.

## Related Issue:
- Part of #643

### DIFF
--- a/tests/unauthorized/filesTest.js
+++ b/tests/unauthorized/filesTest.js
@@ -7,7 +7,14 @@ describe('Unauthorized: Currently testing files management,', function () {
 
   // PACT setup
   const Pact = require('@pact-foundation/pact-web')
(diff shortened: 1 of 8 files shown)


tests/unauthorized/filesTest.js
```diff
@@ -130,7 +139,7 @@ describe('Unauthorized: Currently testing files management,', function () {
   it('checking method : delete', function (done) {
     const newFolder = config.testFolder + '/' + 'new folder'
 
-    oc.files.mkdir(newFolder).then(() => {
+    oc.files.delete(newFolder).then(() => {
```
HariBhandari07, line 142: The interaction is to delete a directory, so it makes more sense deleting a directory instead of making a directory

individual-it (approved):

Issue #646, opened 2021-01-05
convert tests to pact V3


individual-it:
sub task of #644
depends on #645 

in the tests we are dealing a lot with XML responses, to be able to create those easily in the interactions and not have to build complex regex expressions we need to migrate to pact V3. by that we will be able to use `XmlBuilder` and `MatchersV3` see https://github.com/pact-foundation/pact-js/#pact-js-v3

POC: #637

dpakach:
Done with https://github.com/owncloud/owncloud-sdk/issues/645#issuecomment-760682335

Issue #645, opened 2021-01-05
convert tests to jests


individual-it:
sub task of #644 
depends on #643

karma is running in the browser and so it cannot start the provider side tests easily, to overcome the limitations of karma and as preparation for converting the tests to PactV3 we want to migrate the tests to jest

POC: #637

dpakach:
###  Tests to be move to jest
-  [ ] -unauthenticated
    -  [ ] - appTest.js
    -  [ ] - capabilitiesTest.js
    -  [ ] - filesTest.js
    -  [ ] - groupTest.js
    -  [ ] - sharingTest.js
    -  [ ] - userTest.js

-  [ ] -unauthorized 
    -  [ ] - appTest.js 
    -  [ ] - capabilitiesTest.js  
    -  [ ] - filesTest.js  
    -  [ ] - groupTest.js 
    -  [ ] - sharingTest.js 
    -  [ ] - userTest.js 

-  [ ] - appTest.js 
-  [ ] - capabilitiesTest.js
-  [ ] - filesTest.js 
-  [ ] - filesTrashTest.js
-  [ ] - fileVersionTest.js
-  [ ] - groupTest.js
-  [ ] - loginTest.js
-  [ ] - publickLinkTest.js 
-  [ ] - requestsOcsTest.js
-  [ ] - shareRecipientTest.js
-  [ ] - sharingTest.js
-  [ ] - sharingWithAttributesTest.js
-  [ ] - signedUrlIntegrationTest.js
-  [ ] - userTest.js
-  [ ] - xmlParserTest.js
-  [ ] - signUrlTest.js

PR #687, opened 2021-01-19
[Tests-Only] Move unauthenticated tests to jest

saw-jan:
Move unauthenticated tests to jest
 - appsTest.js
 - capabilitiesTest.js
 - filesTest.js
 - groupTest.js
 - sharingTest.js
 - userTest.js

part of #645 & #646

### DIFF
--- a/src/helperFunctions.js
+++ b/src/helperFunctions.js
@@ -295,9 +295,6 @@ class helpers {
       err = 'Please specify an authorization first.'
     }
 
-    if (err) {
-      throw new Error(err)
-    }
     const headers = {
       authorization: this._authHeader,
       'Content-Type': 'application/x-www-form-urlencoded'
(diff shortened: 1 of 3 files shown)


dpakach (commented):
LGTM :+1:

src/helperFunctions.js
```diff
@@ -295,9 +295,6 @@ class helpers {
       err = 'Please specify an authorization first.'
     }
 
-    if (err) {
-      throw new Error(err)
-    }
```
individual-it, line 300: why deleting this?

individual-it (changes requested):

src/helperFunctions.js
saw-jan: @dipak has better explanation.
IMO, Promise will reject if there is any error i.e. `err` and that `if` block was blocking any GET requests made without auth-header which were needed for unauthenticated tests.

individual-it: it was added here https://github.com/owncloud/owncloud-sdk/pull/637/files#diff-b58c3c88931c8c80492c58e8b2f7f2000efc13c8eb29120bbdcc033f3f815873R298-R300 and is not needed

individual-it (approved):

PR #688, opened 2021-01-19
[Tests-Only] move signedUrl, sharingWithAttributres and shareRecipientTest to jest and pactV3

dpakach:
move tests to jest
- signedUrl
- sharingWithAttributes
- shareReciepientTest
- signedUrlIntegrationTest
- XmlParserTest

part of #645 & #646

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -162,7 +162,12 @@ module.exports = {
     '**/tests/unauthorized/appsTest.js',
     '**/tests/unauthorized/capabilitiesTest.js',
     '**/tests/unauthorized/groupTest.js',
-    '**/tests/loginTest.js'
+    '**/tests/loginTest.js',
+    '**/tests/sharingWithAttributesTest.js',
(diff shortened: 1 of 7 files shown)


dpakach:
We get some warnings while running tests now, courtesy of https://github.com/pact-foundation/pact-js/issues/397

individual-it (changes requested):
the problem with https://github.com/pact-foundation/pact-js/issues/577 should be only in the resquest when using formdata, is there also a problem in the response?

jasson99 (commented):
LGTM

dpakach:
> the problem with [pact-foundation/pact-js#577](https://github.com/pact-foundation/pact-js/issues/577) should be only in the resquest when using formdata, is there also a problem in the response?

@individual-it that is existent in any string like body, so that error should occur on `text/plain` responses too

individual-it (approved):

PR #693, opened 2021-01-20
[Tests-Only] move LoginTest to jest and pactv3

dpakach:
move test to jest and pactV3
- LoginTest
 
part of #645 & #646

### DIFF
--- a/tests/loginTest.js
+++ b/tests/loginTest.js
@@ -12,17 +12,12 @@ describe('Main: Currently testing Login and initLibrary,', function () {
   var nonExistentUser = 'nonExistentUser' + timeRightNow
 
   // PACT setup
-  const Pact = require('@pact-foundation/pact-web')
-  const provider = new Pact.PactWeb()
-  const { setGeneralInteractions } = require('./pactHelper.js')
-
-  beforeAll(function () {
-    return setGeneralInteractions(provider)
-  })
-
-  afterAll(function () {
-    return provider.removeInteractions()
-  })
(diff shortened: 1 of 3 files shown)


tests/pactHelper.js
```diff
@@ -64,6 +64,11 @@ const applicationFormUrlEncoded = { 'Content-Type': 'application/x-www-form-urle
 const xmlResponseHeaders = {
   'Content-Type': 'text/xml; charset=utf-8'
 }
+const getAuthHeaders = (username, password) => {
+  const header = `${username}:${password}`
+  const buff = Buffer.from(header)
+  return 'Basic ' + buff.toString('base64')
+}
```
individual-it, line 71: should be reused in more places, can be done in a separate PR

tests/loginTest.js
```diff
@@ -49,63 +44,72 @@ describe('Main: Currently testing Login and initLibrary,', function () {
...
+    await capabilitiesGETRequestValidAuth(provider)
+    await GETRequestToCloudUserEndpoint(provider)
+    return provider.executeTest(async () => {
+      oc = new OwnCloud({
+        baseUrl: config.owncloudURL,
+        auth: {
+          basic: {
+            username: config.username,
+            password: config.password
+          }
```
individual-it, line 104: we can use `createOwncloud` from the Helper, but as it was like that before, it can be done in a separate PR

individual-it (approved):

PR #694, opened 2021-01-20
[Tests-Only] Convert unauthorized userTest to jest

swoichha:
move unauthorized userTests to jest

part of #645 & #646

### DIFF
--- a/tests/unauthorized/userTest.js
+++ b/tests/unauthorized/userTest.js
@@ -15,7 +15,7 @@ describe('Unauthorized: Currently testing user management,', function () {
     createProvider
   } = require('../pactHelper.js')
 
-  const request = async (provider, requestName, method, path) => {
+  const request = async (provider, requestName, method, path, query) => {
     return provider
       .uponReceiving(requestName)
       .withRequest({
@@ -23,7 +23,8 @@ describe('Unauthorized: Currently testing user management,', function () {
         path: path,
         headers: {
           authorization: invalidAuthHeader
-        }
+        },
+        query: query
       })
       .willRespondWith({
         status: 401,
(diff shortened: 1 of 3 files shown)


tests/unauthorized/userTest.js
```diff
@@ -1,287 +1,365 @@
...
+  const {
+    unauthorizedXmlResponseBody,
+    createProvider
+  } = require('../pactHelper.js')
 
-  const request = function (requestName, method, path) {
-    return {
-      uponReceiving: requestName,
-      withRequest: {
+  const request = async (provider, requestName, method, path, query) => {
```
dpakach, line 18:
```suggestion
  const request = (provider, requestName, method, path, query) => {
```

tests/unauthorized/userTest.js
```diff
@@ -1,287 +1,365 @@
...
+          authorization: invalidAuthHeader
+        },
+        query: query
+      })
+      .willRespondWith({
         status: 401,
         headers: {
-          'Content-Type': 'text/xml; charset=utf-8',
-          'Access-Control-Allow-Origin': origin
+          'Content-Type': 'text/xml; charset=utf-8'
```
dpakach, line 32: import this application/xml header from `pactHelper`

tests/unauthorized/userTest.js
```diff
@@ -1,287 +1,365 @@
...
-    }
+      })
   }
 
-  const usersEndpointPath = Pact.Matchers.term({
-    matcher: '.*\\/ocs\\/v1\\.php\\/cloud\\/users$',
-    generate: '/ocs/v1.php/cloud/users'
-  })
+  const adminUserEndpointPath = MatchersV3.regex(
+    '.*\\/ocs\\/v1\\.php\\/cloud\\/users\\/admin$',
```
dpakach, line 39: seems like the endpoint is always same. why use regex here?

tests/unauthorized/userTest.js
```diff
@@ -1,287 +1,365 @@
...
+    '/ocs/v1.php/cloud/users/admin'
+  )
 
-  const groupsEndpointPath = Pact.Matchers.term({
-    matcher: '.*\\/ocs\\/v1\\.php\\/cloud\\/users\\/' + config.testUser + '\\/groups$',
-    generate: '/ocs/v1.php/cloud/users/' + config.testUser + '/groups'
-  })
+  const usersEndpointPath = MatchersV3.regex(
```
dpakach, line 43: same here I dont't think this requires regex

dpakach: also do same for other similar examples

tests/unauthorized/userTest.js
dpakach: also the functionname request is a bit too generic, please change it to something like `addInvalidAuthInteraction`

individual-it (approved):

dpakach (commented):
Was in review, but most of these things apply to many other files too so we can do it som other time

PR #696, opened 2021-01-21
[Tests-Only] move FilesTest to pactV3 and jest

dpakach:
move tests to jest
- FilesTest.js

part of #645 & #646

### DIFF
--- a/src/helperFunctions.js
+++ b/src/helperFunctions.js
@@ -310,7 +310,6 @@ class helpers {
         headers: headers
       })
         .then(async res => {
-          res.headers = new fetch.Headers(res.headers)
           res.statusCode = res.status
           resolve({
             response: res,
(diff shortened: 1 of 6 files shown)


src/helperFunctions.js
```diff
@@ -310,7 +310,7 @@ class helpers {
         headers: headers
       })
         .then(async res => {
-          res.headers = new fetch.Headers(res.headers)
+          // res.headers = new fetch.Headers(res.headers)
```
dpakach, line 313: This causes some tests to fail, Why was this added @individual-it ?

individual-it: I think this is to have the headers as an object not array. What happens with that now? I guess we have an incompatibility between `xmlhttprequest` and `cross-fetch`

individual-it: this line was added in #636 but it seems not to be needed, tests pass at the stage of #636 without it and also pass now without it
CC @LukasHirt

individual-it (approved):

PR #699, opened 2021-01-25
[Tests-Only] move sharing test to pactV3 and jest

dpakach:
move tests to jest
- sharingTest

part of #645 & #646

### DIFF
--- a/tests/sharingTest.js
+++ b/tests/sharingTest.js
@@ -2,6 +2,17 @@ import { MatchersV3, XmlBuilder } from '@pact-foundation/pact/v3'
 
 describe('Main: Currently testing file/folder sharing,', function () {
   const config = require('./config/config.json')
+  config.testFiles = [
+    'test space and + and #.txt',
+    'test.txt',
+    'hello world.txt'
+  ]
+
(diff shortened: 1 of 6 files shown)


tests/sharingTest.js
```diff
@@ -1,26 +1,32 @@
...
+    'test space and + and #.txt',
+    'test.txt',
+    'hello world.txt'
+  ]
+
+  config.testFilesPath = [
+    '%2Ftest+space+and+%2B+and+%23.txt',
+    '%2Ftest.txt',
+    '%2Fhello+world.txt'
+  ]
```
dpakach, line 15: I added this change because with pact when we add an interaction with a query for `'文件.txt'` it doesn't work.

@individual-it I think it is a bug with pact. do you have any idea.

individual-it: can you try to reproduce within the pact-js examples (v3 branch) https://github.com/pact-foundation/pact-js/tree/feat/v3.0.0/examples

dpakach: I've added comments with links to the issue

individual-it (approved):

PR #700, opened 2021-01-26
[Tests-Only] move fileTrashTest to pactv3 and jest

dpakach:
move tests to jest
- FilesTrashTest.js

part of #645 & #646


swoichha (commented):
LGTM

individual-it (changes requested):
please make comments about why tests are skipped and open issues to fix those skips

individual-it (approved):

PR #704, opened 2021-01-27
[tests-only] fix tests that use x-www-form-urlencoded body

individual-it:
these tests have been incomplete and the body part was ignored because of a bug in pact-js see https://github.com/pact-foundation/pact-js/issues/577
as the bug is fixed now, we can bring back these bodies

note: the issue in `signedUrlIntegrationTest.js` seems to be an other one, looking into that now

### DIFF
--- a/tests/sharingTest.js
+++ b/tests/sharingTest.js
@@ -19,6 +19,7 @@ describe('Main: Currently testing file/folder sharing,', function () {
 
   const {
     applicationXmlResponseHeaders,
+    applicationFormUrlEncoded,
     accessControlAllowHeaders,
     accessControlAllowMethods,
     validAuthHeaders,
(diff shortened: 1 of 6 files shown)


tests/sharingTest.js
```diff
@@ -627,10 +630,11 @@ describe('Main: Currently testing file/folder sharing,', function () {
...
             ),
-            headers: validAuthHeaders
-            // TODO: uncomment this line once the issue is fixed
-            // https://github.com/pact-foundation/pact-js/issues/577
-            // body: 'shareType=0&shareWith=' + config.testUser + '&path=' + config.testFilesPath[i] + '&expireDate=' + config.expirationDate
+            headers: {
+              ...validAuthHeaders,
+              ...applicationFormUrlEncoded
+            },
+            body: 'shareType=0&shareWith=' + config.testUser + '&path=' + file + '&expireDate=' + config.expirationDate
```
individual-it, line 637: this function is now called through a `map` now and does not have a loop of its own, so I had to change the path

tests/sharingTest.js
```diff
@@ -848,12 +852,7 @@ describe('Main: Currently testing file/folder sharing,', function () {
         await capabilitiesGETRequestValidAuth(provider)
         await GETRequestToCloudUserEndpoint(provider)
 
-        // TODO: uncomment this line once the issue is fixed
-        // https://github.com/pact-foundation/pact-js/issues/577
-        // await Promise.all(testFiles.map(file => sharePOSTRequestWithExpirationDateSet(provider, file)))
-
-        await sharePOSTRequestWithExpirationDateSet(provider, testFiles[0])
-
+        await Promise.all(config.testFilesPath.map(file => sharePOSTRequestWithExpirationDateSet(provider, file)))
```
individual-it, line 855: `testFilesPath` contains the encoded file list, that matches what the consumer sends

dpakach (approved):
LGTM :+1:

PR #715, opened 2021-02-03
[tests-only] adjust contract of filesTests to match provider

individual-it:


individual-it:
closing in favour of #725

Issue #709, opened 2021-02-01
adjust pact interactions so that they also pass provider side (oc10)


individual-it:
follow up of #662
sub task of #644

Currently we run all tests of the consumer pact against the oc10 provider but most of them fail, see #662 because the contract from the consumer side is not correct or missing given steps.
All failing interactions need to be adjusted, so that the provider accepts them.

saw-jan:
Tests to be adjusted

 - [ ] unauthenticated
   - [ ] appTest.js
   - [ ] capabilitiesTest.js
   - [ ] filesTest.js
   - [ ] groupTest.js
   - [ ] sharingTest.js
   - [ ] userTest.js
 
- [ ] unauthorized
  - [ ] appTest.js
  - [ ] capabilitiesTest.js
  - [ ] filesTest.js
  - [ ] groupTest.js
  - [ ] sharingTest.js
  - [ ] userTest.js
  - [ ] appTest.js

 - [ ] capabilitiesTest.js

 - [ ] filesTest.js wip @dpakach

 - [ ] filesTrashTest.js

 - [ ] fileVersionTest.js

 - [ ] groupTest.js

 - [ ] loginTest.js

 - [ ] publickLinkTest.js

 - [ ] requestsOcsTest.js

 - [ ] shareRecipientTest.js

 - [ ] sharingTest.js

 - [ ] sharingWithAttributesTest.js wip

 - [ ] signedUrlIntegrationTest.js

 - [ ] signUrlTest.js

 - [ ] userTest.js

 - [ ] xmlParserTest.js

PR #716, opened 2021-02-05
[tests-only] some fixes to make more files tests pass in the provider

individual-it:
1. get rid of unneeded CORS headers
2. use matchers for etag and file ids
3. create states for creating files & folders

this is part of #709


saw-jan:
LGTM :+1:

dpakach (approved):
LGTM :+1:

PR #720, opened 2021-02-09
adjust create/delete folder tests to pass on oc10 provider

individual-it:
In order to be able to test the interactions that create or delete folders this PR makes some general changes
- renaming authHeader variables to make it clear that those are admin auth
- new state to recreate (delete and create again) a user
- runs create/delete folder tests as `testUser`
- recreates the `testUser` in provider tests


saw-jan:
LGTM :+1:

dpakach (approved):
LGTM :+1:

PR #721, opened 2021-02-09
update pact-js to 10.0.0-beta.30

individual-it:
a new version of pact-js was published
Features

    fix file locking, add 'overwrite' and 'callbackTimeout' flags (e891fcc), closes #599 #600

Fixes and Improvements

    make the callback timeout configurable with a 5 sec default (a0f0876)


swoichha (approved):
LGTM

PR #726, opened 2021-02-11
revert hacks that were introduced because of char encoding bug in pact

individual-it:
- update pact-js to 10.0.0-beta.31
- remove hacks that were added because of char encoding bug in pact
  see https://github.com/pact-foundation/pact-js/issues/589


dpakach (approved):
LGTM :+1:

Issue #644, opened 2021-01-05
run pact.io provider side tests


individual-it:
all tests have been converted to pact.io see #500 now we want to use the same contracts to test the provider (oc10 & ocis)
for that we need:
- [x] verify consumer side tests #643
- [x] convert the tests to jest - sub task: #645 - POC in #637
- [x] convert tests to use pactV3 - sub task: #646 - POC in #637
- [x] run provider tests with oc10 in drone - sub task: #647 - POC in #662
- [x] adjust pact interactions so that they also pass provider side (oc10) #709
- [x] run provider tests with ocis in drone - sub task: #648 - done in  #740
- [x] adjust pact interactions so that they also pass provider side (ocis) #747
- [x] adjust interaction for interactions unskipped in https://github.com/owncloud/owncloud-sdk/pull/735

PR #728, opened 2021-02-12
[tests-only] Adjust pact interaction of sharingTest

saw-jan:
part of https://github.com/owncloud/owncloud-sdk/issues/709 and https://github.com/owncloud/owncloud-sdk/issues/644

adjustment for test suites:
- sharingTest

### DIFF
--- a/tests/helpers/sharingHelper.js
+++ b/tests/helpers/sharingHelper.js
@@ -4,6 +4,7 @@ const {
   applicationFormUrlEncoded,
   sanitizeUrl
 } = require('../pactHelper.js')
+const config = require('../config/config.json')
 
 const shareEndPoint = '/ocs/v1.php/apps/files_sharing/api/v1/shares'
 const publicFilesEndPoint = '/remote.php/dav/public-files'
@@ -38,7 +39,7 @@ const getPublicFilesEndPoint = function () {
  * @param {string} name
  * @param {boolean} publicUpload
  * @param {string} password
- * @param {Date} expireDate
+ * @param {string} expireDate
  * @returns {*} result of the fetch request
  */
 const shareResource = function (
@@ -91,29 +92,18 @@ const getShareInfoByPath = function (username, userPassword, path) {
 const validateParams = function (data) {
   const params = new URLSearchParams()
   if (data) {
-    if (data.path) {
-      params.append('path', data.path)
-    }
-    if (data.shareType !== undefined && data.shareType !== null && data.shareType !== '') {
-      params.append('shareType', data.shareType)
-    }
-    if (data.shareWith) {
-      params.append('shareWith', data.shareWith)
-    }
-    if (data.permissions) {
-      params.append('permissions', data.permissions)
-    }
-    if (data.name) {
-      params.append('name', data.name)
-    }
-    if (data.password) {
-      params.append('password', data.password)
-    }
-    if (data.expireDate) {
-      params.append('expireDate', data.expireDate)
-    }
-    if (data.publicUpload !== undefined && data.publicUpload !== null && data.publicUpload !== '') {
-      params.append('publicUpload', data.publicUpload.toString().toLowerCase())
+    for (const key in data) {
+      let value = data[key]
+      if (key === 'publicUpload' || key === 'shareType') {
+        if (value !== undefined && value !== null && value !== '') {
+          value = value.toString().toLowerCase()
+          params.append(key, value)
+        }
+      } else {
+        if (value) {
+          params.append(key, value)
+        }
+      }
     }
   }
 
@@ -147,9 +137,48 @@ const createFileInLastPublicShare = function (token, fileName) {
   })
 }
 
+/**
+ * convert JS object to form url encoded data.
+ *
+ * @param {object} obj
+ * @returns {string}
+ */
+const toFormUrlEncoded = function (obj) {
+  if (obj) {
+    for (const key in obj) {
+      let value = obj[key]
+      if (key === 'publicUpload') {
+        value = obj[key].toString().toLowerCase()
+      }
+      return `${key}=${value}`
+    }
+  }
+}
+
+/**
+ * @param {sring} resource
+ * @param {sring} resourceType
+ *
+ * @returns {object} shareId and shareToken
+ */
+const getShareIdToken = (resource, resourceType) => {
+  let shareId, shareToken
+  if (resourceType === 'folder') {
+    shareId = config.testFolderId
+    shareToken = config.shareTokenOfPublicLinkFolder
+  } else {
+    const iteration = config.testFiles.indexOf(resource)
+    shareId = config.testFilesId[iteration]
+    shareToken = config.testFilesToken[iteration]
+  }
+  return { shareId, shareToken }
+}
+
 module.exports = {
   shareResource,
   getShareInfoByPath,
   createFolderInLastPublicShare,
-  createFileInLastPublicShare
+  createFileInLastPublicShare,
+  toFormUrlEncoded,
+  getShareIdToken
 }
--- a/tests/sharingTest.js
+++ b/tests/sharingTest.js
@@ -1,33 +1,52 @@
 import { MatchersV3, XmlBuilder } from '@pact-foundation/pact/v3'
 
+const { fromProviderState } = MatchersV3
+const {
+  toFormUrlEncoded,
+  getShareIdToken
+} = require('./helpers/sharingHelper')
+
+const {
+  givenUserExists,
+  givenGroupExists,
+  givenShareExists,
+  givenFileFolderIsCreated,
+  givenResourceIsShared
+} = require('./helpers/providerStateHelper')
+
 describe('Main: Currently testing file/folder sharing,', function () {
   const config = require('./config/config.json')
-  const username = config.adminUsername
 
   const {
     applicationXmlResponseHeaders,
     applicationFormUrlEncoded,
-    accessControlAllowHeaders,
-    accessControlAllowMethods,
-    validAdminAuthHeaders,
-    xmlResponseHeaders,
+    htmlResponseHeaders,
     getCapabilitiesInteraction,
     getCurrentUserInformationInteraction,
     createOwncloud,
     ocsMeta,
     shareResponseOcsData,
     createProvider,
-    origin
+    getAuthHeaders
   } = require('./pactHelper.js')
 
   // TESTING CONFIGS
   const testUserPassword = config.testUserPassword
-  const testUser = config.testUser
   const testGroup = config.testGroup
-  const testFolder = '/' + config.testFolder
+  const testFolder = config.testFolder
   const nonExistentFile = config.nonExistentFile
   const expirationDate = config.expirationDate
   const testFiles = config.testFiles
+  const testFilesId = config.testFilesId
+
+  const sharer = config.testUser
+  const sharerPassword = config.testUserPassword
+  const sharee = config.testUser2
+  const shareePassword = config.testUser2Password
+
+  const validAuthHeaders = {
+    authorization: getAuthHeaders(sharer, sharerPassword)
+  }
 
   // CREATED SHARES
   const sharedFiles = {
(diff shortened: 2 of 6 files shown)


tests/sharingTest.js
```diff
@@ -35,175 +37,480 @@ describe('Main: Currently testing file/folder sharing,', function () {
     '文件.txt': 19
   }
   const testFolderShareID = 9
+  const testFolderShareToken = '6Bvt5fDajGxiV3P'
 
   // CONSTANTS FROM lib/public/constants.php
   const OCS_PERMISSION_READ = 1
   const OCS_PERMISSION_UPDATE = 2
-  const OCS_PERMISSION_CREATE = 4
+  // const OCS_PERMISSION_CREATE = 4
```
jasson99, line 45: Can this be removed then?

tests/sharingTest.js
```diff
@@ -35,175 +37,480 @@ describe('Main: Currently testing file/folder sharing,', function () {
...
+      await givenUserExists(provider, testUser, testUserPassword)
+      return givenResourceShared(username, password, provider, 0, resource, testUser)
+    } else if (shareType === 1) {
+      await givenGroupExists(provider, testGroup)
+      return givenResourceShared(username, password, provider, 1, resource, testGroup)
+    }
+  }
+  const initFileFolderUserGroupShare = async (provider, shareType, resource, resourceType) => {
+    await createFileOrFolder(adminUser, adminPassword, provider, resource, resourceType)
+    return createShare(adminUser, adminPassword, provider, shareType, resource)
```
jasson99, line 94: These should be performed by test user rather than admin user

jasson99 (changes requested):
The requests should be performed by test user rather than admin user

tests/sharingTest.js
```diff
@@ -1,224 +1,538 @@
...
   const testFiles = config.testFiles
 
+  const sharer = config.testUser
+  const sharerPassword = config.testUserPassword
+  const sharee = config.testUser2
+  const shareePassword = config.testUser2Password
+
+  const validAuthHeaders = {
+    authorization: getAuthHeaders(sharer, sharerPassword)
+  }
```
jasson99, line 39: may be we can shift this in pactHelper and can use in other files as well, as most of them require validUserAuthHeaders now.

jasson99 (commented):
minor suggestions only

tests/sharingTest.js
saw-jan: we can implement something like this
```js
const getValidUserAuthHeaders = (username, password) => ({
  authorization: getAuthHeaders(username, password)
})
```
and use it like
```js
// on every interaction
...
headers: getValidUserAuthHeaders(username, password)
...

OR

validUserAuthHeaders = getValidUserAuthHeaders(username, password)
...
headers: validUserAuthHeaders
...
```

It is also possible to introduce some confusion between `getAuthHeaders` and `getValidUserAuthHeaders`, and workarounds

tests/providerTest.js
```diff
@@ -144,6 +149,32 @@ describe('provider testing', () => {
...
       'provider base url is returned': () => {
         return { providerBaseURL: process.env.PROVIDER_BASE_URL }
+      },
+      'resource is shared': (setup, parameters) => {
+        if (setup) {
+          const { username, password, resource, shareType, shareWith } = parameters
+          const response = shareResource(username, password, resource, shareType, shareWith)
+
+          const { status } = getOCSMeta(response)
+          /* eslint-disable-next-line */
```
dpakach, line 159: please disable only one rule that's causing error

saw-jan: this is to disable the following one line,
how can I disable a specific rule?

individual-it: e.g `// eslint-disable-next-line no-useless-escape`

tests/helpers/sharingHelper.js
```diff
@@ -147,9 +147,38 @@ const createFileInLastPublicShare = function (token, fileName) {
...
+    }
+    if (obj.password) {
+      toUpdate = `password=${obj.password}`
+    }
+    if (obj.expireDate) {
+      toUpdate = `expireDate=${obj.expireDate}`
+    }
+    if (obj.publicUpload !== undefined && obj.publicUpload !== null && obj.publicUpload !== '') {
+      toUpdate = `publicUpload=${obj.publicUpload.toString().toLowerCase()}`
+    }
```
dpakach, line 173: how about just loop through the object and append, to make it more cleaner

tests/sharingTest.js
```diff
@@ -1,33 +1,41 @@
...
+  const testFilesId = config.testFilesId
+
+  const sharer = config.testUser
+  const sharerPassword = config.testUserPassword
+  const sharee = config.testUser2
+  const shareePassword = config.testUser2Password
+
+  const validAuthHeaders = {
+    authorization: getAuthHeaders(sharer, sharerPassword)
+  }
```
jasson99, line 38: IMO, this can be used in other files as well, so we can keep this in pact helper directly, with getAuthHeaders, validAdminAuthHeaders. Using it this way can be uniform when we can use in other files as well.

jasson99 (commented):
Just minor suggestion. Other than that, Looks good to me

tests/sharingTest.js
saw-jan: we can make it at once for all tests in another PR

swoichha (commented):
lgtm

tests/helpers/sharingHelper.js
```diff
@@ -147,9 +136,31 @@ const createFileInLastPublicShare = function (token, fileName) {
...
 }
 
+/**
+ * convert JS object to form url encoded data.
+ *
+ * @param {object} obj
+ * @returns {string}
+ */
+const toFormUrlEncoded = function (obj) {
+  let toUpdate
```
individual-it, line 146: that variable name somehow does not have anything to do with the function name. maybe call it something like `formData`

Talank (commented):
LGTM

tests/sharingTest.js
```diff
@@ -36,190 +44,494 @@ describe('Main: Currently testing file/folder sharing,', function () {
...
+
+  describe('Currently testing folder sharing,', function () {
+    describe('sharedFolderByLink,', function () {
+      describe('checking the shared folder,', function () {
+        it('checking method : isShared with shared folder', async function () {
+          const provider = createProvider()
+          await getCapabilitiesInteraction(provider, sharer, sharerPassword)
+          await getCurrentUserInformationInteraction(provider, sharer, sharerPassword)
+          await getIsSharedInteraction(provider, `check whether '${testFolder}' is shared or not (public link share)`, 3, testFolder, 'folder')
```
individual-it, line 272: pretty long line

tests/sharingTest.js
```diff
@@ -36,190 +44,494 @@ describe('Main: Currently testing file/folder sharing,', function () {
...
+  const givenFolderExists = (provider, username, password, resource) => {
+    return provider
+      .given('folder exists', { username, password, folderName: resource })
+  }
+  const givenResourceShared = async (provider, username, password, shareType, resource, shareWith) => {
+    return provider
+      .given('resource is shared', { username, password, resource, shareType, shareWith })
+  }
+  let shareId, shareToken
+  const createFileOrFolder = (provider, username, password, resource, resourceType) => {
```
individual-it, line 75: lets be consistent with the naming
```suggestion
  const givenfFileFolderExists = (provider, username, password, resource, resourceType) => {
```

tests/sharingTest.js
```diff
@@ -36,190 +44,494 @@ describe('Main: Currently testing file/folder sharing,', function () {
...
+      shareToken = testFolderShareToken
+      return givenFolderExists(provider, username, password, resource)
+    } else {
+      const iteration = testFiles.indexOf(resource)
+      shareId = config.testFilesId[iteration]
+      shareToken = config.testFilesToken[iteration]
+      return givenFileExists(provider, username, password, resource)
+    }
+  }
+  const createShare = async (provider, username, password, shareType, resource) => {
```
individual-it, line 87: make a consistent name with `given`

tests/sharingTest.js
```diff
@@ -36,190 +44,494 @@ describe('Main: Currently testing file/folder sharing,', function () {
...
+        it('adding password', async function () {
+          const toUpdate = { password: '1234' }
+          const additionalBodyElement = {
+            share_with: testGroup,
+            share_with_displayname: testGroup
+          }
+          const provider = createProvider()
+          await getCapabilitiesInteraction(provider, sharer, sharerPassword)
+          await getCurrentUserInformationInteraction(provider, sharer, sharerPassword)
+          await updateShareInteraction(provider, '(group share: set share password)', 1, testFolder, 'folder', toUpdate, additionalBodyElement)
```
individual-it, line 487: too long line

tests/sharingTest.js
```diff
@@ -710,19 +956,21 @@ describe('Main: Currently testing file/folder sharing,', function () {
...
-        await shareGETExistingNonSharedFile(provider)
+        await shareGETExistingNonSharedFile(provider, '', fileName)
 
         return provider.executeTest(async () => {
-          const oc = createOwncloud()
+          const oc = createOwncloud(sharer, sharerPassword)
           await oc.login()
-          const suffix = '123'
-          return oc.files.putFileContents('newFileCreated' + suffix).then(status => {
+          return oc.files.putFileContents(fileName).then(status => {
```
individual-it, line 971: we don't need to write the file here, because this is not about putFileContents but about testing the consumer side of `isShared`
but let's leave it as it is here for now

individual-it (changes requested):

tests/helpers/sharingHelper.js
```diff
@@ -147,9 +136,31 @@ const createFileInLastPublicShare = function (token, fileName) {
   })
 }
 
+/**
+ * convert JS object to form url encoded data.
+ *
+ * @param {object} obj
+ * @returns {string}
+ */
+const toFormUrlEncoded = function (obj) {
```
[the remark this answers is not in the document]
saw-jan: it is mainly used to convert a share's update parameter which is in js object into `key=value` pair
As only one update parameter works per request, I made a break

individual-it (approved):

tests/sharingTest.js
```diff
@@ -408,83 +736,71 @@ describe('Main: Currently testing file/folder sharing,', function () {
...
-          for (file in sharedFiles) {
-            await oc.shares.getShares(file).then(shares => {
-              expect(shares.constructor).toBe(Array)
-              let flag = 0
-              for (let i = 0; i < shares.length; i++) {
-                const share = shares[i]
-                if (allIDs.indexOf(share.getId()) > -1) {
-                  flag = 1
-                }
-              }
```
dpakach, line 458: why did you remove this part of the test

tests/sharingTest.js
```diff
@@ -332,71 +685,46 @@ describe('Main: Currently testing file/folder sharing,', function () {
...
                 expect(typeof (share)).toBe('object')
                 expect(share.getId()).toBeGreaterThan(-1)
               }).catch(error => {
                 expect(error).toBe(null)
               })
-            }
-          })
-        })
-
-        it('getShares for shared file', async function () {
```
dpakach, line 355: why is this test removed?

tests/sharingTest.js
```diff
@@ -232,95 +600,80 @@ describe('Main: Currently testing file/folder sharing,', function () {
           })
         })
 
-        it('checking method : getShares for shared file', async function () {
```
dpakach, line 235: is this test removed?

individual-it: isn't is that here https://github.com/owncloud/owncloud-sdk/pull/728/files#diff-3e96a43471ac474278d7585f7a1ab0a201ce7c5a935710b88c2a9304710d8a76R763 ?

saw-jan: that test is in this line https://github.com/owncloud/owncloud-sdk/blob/a4355e367127b9412421e5d8a193c6499c0757f0/tests/sharingTest.js#L763

tests/sharingTest.js
saw-jan: I directly counted share count and asserted it, because at given state three shares are created

dpakach (approved):
LGTM :+1:

PR #729, opened 2021-02-12
run all tests

individual-it:
just run all test files, no need to list them one-by-one. Some files got forgotten

Now running 251 tests vs 237 before


dpakach (approved):
LGTM :+1:

PR #732, opened 2021-02-15
[Tests-Only] Adjust pact interactions for provider tests for signedUrlIntegration

jasson99:
## Description:
In this PR, the signedUrlIntegration test file is adjusted such that the tests pass on the provider side as well.

## Related Issue: 
- https://github.com/owncloud/owncloud-sdk/issues/709

### DIFF
--- a/tests/webdavHelper.js
+++ b/tests/webdavHelper.js
@@ -114,11 +114,26 @@ const listVersionsFolder = function (user, password, fileId) {
   return listResult.text()
 }
 
+const getSignKey = function (username, password) {
+  const endpoint = process.env.PROVIDER_BASE_URL + '/ocs/v1.php/cloud/user/signing-key?format=json'
+  const response = fetch(endpoint, {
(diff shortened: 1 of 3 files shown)


tests/webdavHelper.js
```diff
@@ -82,8 +82,18 @@ const deleteItem = function (user, password, itemName) {
   })
 }
 
+const getSignKey = function (endpoint, username, password) {
```
dpakach, line 85:
```suggestion
const getSignKey = function (username, password) {
```
endpoint is always same

jasson99: fixed

dpakach (commented):
Lgtm, I dont have much idea about how signed urls work, @individual-it maybe you can review this

individual-it (changes requested):
good job, but downloading a signed url seems not to work yet
image

individual-it (approved):

Issue #712, opened 2021-02-02
replace davclient.js with perry-mitchell/webdav-client


individual-it:
subtask of #705

the current davclient is not working on node, so we need to replace it with a client that works with node as well as browsers.
one possibility is https://github.com/perry-mitchell/webdav-client
It used not to support customer dav props see https://github.com/perry-mitchell/webdav-client/pull/165 (current status need to be researched and if needed an upstream PR need to be created)

dpakach:
Fixed by #735

Issue #705, opened 2021-01-27
Make owncloud-sdk compatible with nodejs


dpakach:
The sdk still uses many browser specific APIs such as `XmlHttpRequest`, `DomParser` etc. All those references should be replaced with libraries compatible with browser and nodejs both to make the sdk fully compatible with nodejs

TODO:
- [ ] - Unskip all tests that are skipped because of this issue

dpakach:
Fixed by https://github.com/owncloud/owncloud-sdk/pull/735

PR #735, opened 2021-02-16
replace owncloud/davclient.js with webdav

dpakach:
### Description
replace owncloud/davclient.js with webdav
Related Issue https://github.com/owncloud/owncloud-sdk/issues/712
Related https://github.com/owncloud/owncloud-sdk/issues/705

This PR replaces `owncloud/davclient.js` with `perry-mitchell/webdav-client`. This makes owncloud-sdk compatible with nodejs.

Changes -
- [x] - Import the code related to making dav request and parsing dav response into new Dav class from https://github.com/owncloud/davclient.js/blob/master/lib/client.js
- [x] - Replace the use of https://github.com/owncloud/davclient.js with the newly created Dav class
- [x] - Unskip and adjust all the tests so that they pass


dpakach:
### TODO
- [ ] - It Still doesn't work with web fully, make all the tests pass on web (https://github.com/owncloud/web/pull/4738)
          - this will probably require code changes on web
  In web folders are redered as files, this may be because the `d:resoucetype` property used to be an array of xmlNode (eg. [d:collection] where `d:collection` is xmlNode), but it now returns it as array of string (eg. ["d:collection"]) this needs to be handled in web. There may also be other errors that needs to be checked.

Screenshot from 2021-02-17 12-45-57

dpakach:
@LukasHirt Looks like everything is working now. I tested it on web https://github.com/owncloud/web/pull/4738
All tests are passing on web. I mostly just copied the code from owncloud/davclient.js you may find some improvements there.
can you review when you have time?

src/dav.js
```diff
@@ -0,0 +1,376 @@
...
+   * @param {Object} [headers] headers
+   * @return {Promise}
+   */
+  propFind (path, properties, depth, headers, options = {}) {
+    if (typeof depth === 'undefined') {
+      depth = '0'
+    }
+
+    // depth header must be a string, in case a number was passed in
+    depth = '' + depth
```
individual-it, line 55: what is a sting is passed but something useless like `very deep`?
or it being an empty sting?

src/dav.js
```diff
@@ -0,0 +1,376 @@
...
+   *
+   * @param {string} url Url to do the propfind request on
+   * @param {Array} properties List of properties to retrieve.
+   * @param {string} depth "0", "1" or "infinity"
+   * @param {Object} [headers] headers
+   * @return {Promise}
+   */
+  propFind (path, properties, depth, headers, options = {}) {
+    if (typeof depth === 'undefined') {
+      depth = '0'
```
individual-it, line 51: why not having the depth & headers parameter having default values?  
`propFind (path, properties, depth = '0', headers = {}, options = {}) {`

src/filesTrash.js
```diff
@@ -46,9 +40,9 @@ class FilesTrash {
     const headers = this.helpers.buildHeaders()
     const target = '/trash-bin/' + this.helpers.getCurrentUser().id + '/' + path
 
-    return this.davClient.propFind(this.helpers._buildFullWebDAVPathV2(target), properties, depth, headers).then(result => {
+    return this.davClient.propFind(this.helpers._buildFullWebDAVPathV2(target), properties, depth, headers, { version: 'v2' }).then(result => {
```
individual-it, line 43: long line

src/filesTrash.js
```diff
@@ -71,11 +65,11 @@ class FilesTrash {
       target += item
     }
 
-    return this.davClient.request('DELETE', this.helpers._buildFullWebDAVPathV2(target), headers).then(result => {
+    return this.davClient.request('DELETE', this.helpers._buildFullWebDAVPathV2(target), headers, null, { version: 'v2' }).then(result => {
```
individual-it, line 68: long line

src/filesTrash.js
```diff
@@ -99,13 +93,13 @@ class FilesTrash {
     const source = '/trash-bin/' + this.helpers.getCurrentUser().id + '/' + fileId
     const target = '/files/' + this.helpers.getCurrentUser().id + '/' + originalLocation
 
-    headers.Destination = this.helpers._buildFullWebDAVPathV2(target)
+    headers.Destination = this.helpers._buildFullWebDAVURLV2(target)
     headers.Overwrite = overWrite ? 'T' : 'F'
-    return this.davClient.request('MOVE', this.helpers._buildFullWebDAVPathV2(source), headers).then(result => {
+    return this.davClient.request('MOVE', this.helpers._buildFullWebDAVPathV2(source), headers, null, { version: 'v2' }).then(result => {
```
individual-it, line 98: long line

src/systemTags.js
```diff
@@ -48,11 +42,11 @@ class SystemTags {
     headers['Content-Type'] = 'application/json'
     const body = JSON.stringify(tagInfo)
 
-    return this.davClient.request('POST', this.helpers._buildFullWebDAVPathV2('systemtags'), headers, body, '').then(result => {
+    return this.davClient.request('POST', this.helpers._buildFullWebDAVPathV2('systemtags'), headers, body, { version: 'v2' }).then(result => {
```
individual-it, line 45: long line

tests/filesTest.js
```diff
@@ -833,24 +849,25 @@ describe('Main: Currently testing files management,', function () {
...
-  describe.skip('TUS detection', function () {
-    const tusSupportRequest = (provider, enabled = true) => {
+  describe('TUS detection', function () {
+    const tusSupportRequest = (provider, enabled = true, path = '/') => {
       let respHeaders = applicationXmlResponseHeaders
       if (enabled) {
         respHeaders = {
           ...respHeaders,
-          'Content-Type': 'application/xml',
+          'Content-Type': 'application/xml; charset=utf-8',
```
individual-it, line 858: isn't there a variable for that?

src/dav.js
```diff
@@ -0,0 +1,376 @@
...
+        })
+        response.propStat.push(propStat)
+      })
+
+      result.push(response)
+    })
+    return result
+  }
+
+  parseClarkNotation (propertyName) {
```
individual-it, line 365: unit tests for this function would be useful to make sure the regex is correct and stays correct

src/dav.js
```diff
@@ -0,0 +1,376 @@
...
+  /**
+   * Generates a MKCOL request.
+   * If attributes are given, it will use an extended MKCOL request.
+   *
+   * @param {string} url Url to do the proppatch request on
+   * @param {Object.<String,String>} [properties] list of properties to store.
+   * @param {Object} [headers] headers
+   * @return {Promise}
+   */
+  mkcol (url, properties, headers) {
```
individual-it, line 187: unit tests for this function would be great, to make sure the composed body is correct.

src/dav.js
```diff
@@ -0,0 +1,376 @@
...
+
+  /**
+   * Generates a propPatch request.
+   *
+   * @param {string} url Url to do the proppatch request on
+   * @param {Object.<String,String>} properties List of properties to store.
+   * @param {Object} [headers] headers
+   * @return {Promise}
+   */
+  propPatch (url, properties, headers) {
```
individual-it, line 152: unit tests for this function would be great, to make sure the composed body is correct.

src/dav.js
```diff
@@ -0,0 +1,376 @@
...
+  /**
+   * Generates a propFind request.
+   *
+   * @param {string} url Url to do the propfind request on
+   * @param {Array} properties List of properties to retrieve.
+   * @param {string} depth "0", "1" or "infinity"
+   * @param {Object} [headers] headers
+   * @return {Promise}
+   */
+  propFind (path, properties, depth, headers, options = {}) {
```
individual-it, line 49: unit tests for this function would be great, to make sure the composed body is correct.

individual-it (changes requested):
wow, what an epic work :rocket: . I'm happy that soo many more tests can be enabled

it would be also interesting to run the acceptance tests of https://github.com/owncloud/web/ with this changes before merging them to make sure it also works correctly in the browser evn.

src/dav.js
dpakach: I think that is covered very well by the pact tests. These methods make HTTP requests if we want to do unit tests we will need to intercept the HTTP request and check but not really necessary IMO

src/dav.js
dpakach: If you pass invalid depth then it will be ignored by the server, so it's ok IMO

src/dav.js
```diff
@@ -0,0 +1,372 @@
...
+    var namespace
+    for (namespace in this.xmlNamespaces) {
+      body += ' xmlns:' + this.xmlNamespaces[namespace] + '="' + namespace + '"'
+    }
+    body += '>\n' +
+          '  <d:prop>\n'
+
+    for (var ii in properties) {
+      // eslint-disable-next-line no-prototype-builtins
+      if (!properties.hasOwnProperty(ii)) {
```
kulmann, line 70:
```suggestion
      if (!Object.prototype.hasOwnProperty.call(properties, ii)) {
```
if you do it like this you can get rid of the eslint disable command

src/dav.js
```diff
@@ -0,0 +1,372 @@
...
+      } else {
+        body += '    <x:' + property.name + ' xmlns:x="' + property.namespace + '" />\n'
+      }
+    }
+    body += '  </d:prop>\n'
+    body += '</d:propfind>'
+
+    return this.request('PROPFIND', path, headers, body, options).then(
+      function (result) {
+        if (depth === '0') {
```
kulmann, line 86: You could assign the depth specific value to a variable and then have simpler return handling:
```js
const body = (depth === '0' ? result.body : result.body[0])
return {
             status: result.status,
             body,
             res: result
}
```

src/dav.js
```diff
@@ -0,0 +1,372 @@
...
+   * @param {Object.<String,String>} properties
+   * @return {String} XML "<d:set>" block
+   */
+  _renderPropSet (properties) {
+    var body = '  <d:set>\n' +
+          '   <d:prop>\n'
+
+    for (var ii in properties) {
+      // eslint-disable-next-line no-prototype-builtins
+      if (!properties.hasOwnProperty(ii)) {
```
kulmann, line 115: See the other hasOwnProperty related comment to make eslint happy.

src/dav.js
```diff
@@ -0,0 +1,372 @@
...
+   * @return {Promise}
+   */
+  propFind (path, properties, depth = '0', headers = {}, options = {}) {
+    // depth header must be a string, in case a number was passed in
+    depth = '' + depth
+
+    headers = headers || {}
+
+    headers.Depth = depth
+    headers['Content-Type'] = 'application/xml; charset=utf-8'
```
kulmann, line 56: You had a constant for the tests with `'application/xml; charset=utf-8'` as content-type headers. Could you move that to another location and reuse it in this class as well (there are other occurrences in this class).

tests/pactHelper.js
```diff
@@ -108,10 +108,22 @@ const webdavExceptionResponseBody = (exception, message) => new XmlBuilder('1.0'
       .appendElement('s:message', '', message)
   })
 
-const webdavPath = resource => MatchersV3.regex(
-  '.*\\/remote\\.php\\/webdav\\/' + webdavMatcherForResource(resource),
-  `/remote.php/webdav/${resource}`
-)
+const webdavPath = (resource, version = 'old') => {
```
kulmann, line 111: Do I get it right, that you could remove the versioning here entirely? Looks like a leftover from the development of this PR. Matchers below are identical.

kulmann (changes requested):

dpakach:
@kulmann I've fixed everything can you review it. Also all tests pass on web https://github.com/owncloud/web/pull/4738

kulmann (approved):
LGTM now. Impressive piece of work! 🚀

dpakach:
Thanks a lot @kulmann, I've rebased, will merge when CI is green

individual-it (changes requested):
there are now suddenly 81 failures in the provider tests, need investigation
image

individual-it:
I thought that if already passing and approved provider tests would start failing CI would fail, but that seems not to be the case

dpakach:
@individual-it I have confirmed that all the failures **on oc10 provider tests** exist on the previously skipped tests.
There was some failure on public link tests which i have fixed on https://github.com/owncloud/owncloud-sdk/pull/735/commits/c024777722794bd004adfb4e368e15004adfe38b
They are now down to 73 warnings. Can you review and merge now?

individual-it (approved):

PR #737, opened 2021-02-16
[tests-only] Adjust pact interaction for publicLinkTest

saw-jan:
part of https://github.com/owncloud/owncloud-sdk/issues/709 and https://github.com/owncloud/owncloud-sdk/issues/644

adjustment for test suites:
- publicLinkTest

### DIFF
--- a/tests/helpers/sharingHelper.js
+++ b/tests/helpers/sharingHelper.js
@@ -1,0 +1,83 @@
+const fetch = require('sync-fetch')
+const {
+  getAuthHeaders,
+  applicationFormUrlEncoded
+} = require('../pactHelper.js')
+
+const shareEndPoint = '/ocs/v1.php/apps/files_sharing/api/v1/shares'
+const publicFilesEndPoint = '/remote.php/dav/public-files'
+
+/**
+ * returns the full and sanitized URL of the dav resource
+ *
+ * @returns {string}
+ */
+const getSharingEndPoint = function () {
+  return (process.env.PROVIDER_BASE_URL + shareEndPoint)
+    .replace(/([^:])\/{2,}/g, '$1/')
+}
+
+/**
+ * returns the full and sanitized URL of public-files
+ *
+ * @returns {string}
+ */
+const getPublicFilesEndPoint = function () {
+  return (process.env.PROVIDER_BASE_URL + publicFilesEndPoint)
+    .replace(/([^:])\/{2,}/g, '$1/')
+}
+
+/**
+ * share a file or folder using webDAV api.
+ *
+ * @param {string} username
+ * @param {string} password
+ * @param {string} path
+ * @param {number} shareType
+ * @param {string} shareWith
+ * @param {number} permissions
+ * @returns {*} result of the fetch request
+ */
+const shareResource = function (username, password, path, shareType, shareWith, permissions) {
+  return fetch(getSharingEndPoint() + '?format=json', {
+    method: 'POST',
+    body: `path=${path}&shareType=${shareType}&sharewith=${shareWith}&permissions=${permissions}`,
+    headers: {
+      authorization: getAuthHeaders(username, password),
+      ...applicationFormUrlEncoded
+    }
+  })
+}
+
+/**
+ * creates folder in the last shared public share
+ *
+ * @param {string} token
+ * @param {string} folderName
+ * @returns {*} result of the fetch request
+ */
+const createFolderInLastPublicShare = function (token, folderName) {
+  return fetch(getPublicFilesEndPoint() + `/${token}/${folderName}?format=json`, {
+    method: 'MKCOL'
+  })
+}
+
+/**
+ * creates file in the last shared public share
+ *
+ * @param {string} token
+ * @param {string} fileName
+ * @returns {*} result of the fetch request
+ */
+const createFileInLastPublicShare = function (token, fileName) {
+  return fetch(getPublicFilesEndPoint() + `/${token}/${fileName}?format=json`, {
+    method: 'PUT',
+    body: 'a file'
+  })
+}
+
+module.exports = {
+  shareResource,
+  createFolderInLastPublicShare,
+  createFileInLastPublicShare
+}
--- a/tests/helpers/sharingHelper.js
+++ b/tests/helpers/sharingHelper.js
@@ -1,7 +1,8 @@
 const fetch = require('sync-fetch')
 const {
   getAuthHeaders,
-  applicationFormUrlEncoded
+  applicationFormUrlEncoded,
+  sanitizeUrl
 } = require('../pactHelper.js')
 
 const shareEndPoint = '/ocs/v1.php/apps/files_sharing/api/v1/shares'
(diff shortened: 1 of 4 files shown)


tests/publicLinkTest.js
```diff
@@ -5,6 +5,9 @@ import { MatchersV3, XmlBuilder } from '@pact-foundation/pact/v3'
 describe('oc.publicFiles', function () {
   const config = require('./config/config.json')
   const using = require('jasmine-data-provider')
+  const { adminUsername } = config
+  const { testUser, testUserPassword } = config
+  const { testFolder, testFile } = config
```
kiranparajuli589, line 10:
```suggestion
  const { adminUsername, testUser, testUserPassword, testFolder, testFile } = config
```

tests/publicLinkTest.js
```diff
@@ -49,11 +52,14 @@ describe('oc.publicFiles', function () {
...
+      .given('the user is recreated', { username: testUser, password: testUserPassword })
+      .given('folder exists', { username: testUser, password: testUserPassword, folderName: testFolder })
+      .given('resource is shared', { username: testUser, password: testUserPassword, resource: testFolder, shareType: 3, permissions: 15 })
       .uponReceiving(description)
       .withRequest({
         method: method,
-        path: MatchersV3.regex(
-          `.*\\/remote\\.php\\/dav\\/public-files\\/${config.shareTokenOfPublicLinkFolder}\\/foo`,
+        path: MatchersV3.fromProviderState(
+          '/remote.php/dav/public-files/${token}/foo', /* eslint-disable-line no-template-curly-in-string */
```
kiranparajuli589, line 62: does the var `token` get replaced this way? why not using ``backtick(`)``?

saw-jan: Yeah, that's how the `fromProviderState` works.

kiranparajuli589 (commented):
Lgtm ✌️

tests/publicLinkTest.js
```diff
@@ -49,11 +50,14 @@ describe('oc.publicFiles', function () {
 
   const folderInPublicShareInteraction = (provider, description, method, statusCode) => {
     return provider
+      .given('the user is recreated', { username: testUser, password: testUserPassword })
+      .given('folder exists', { username: testUser, password: testUserPassword, folderName: testFolder })
+      .given('resource is shared', { username: testUser, password: testUserPassword, resource: testFolder, shareType: 3, permissions: 15 })
```
individual-it, line 55: too long line, simmilar in other places

tests/publicLinkTest.js
```diff
@@ -212,7 +220,7 @@ describe('oc.publicFiles', function () {
                 expect(files[0].getName()).toBe(testFolderShare.getToken())
                 expect(files[0].getPath()).toBe('/')
                 expect(files[0].getProperty(oc.publicFiles.PUBLIC_LINK_ITEM_TYPE)).toBe('folder')
-                expect(files[0].getProperty(oc.publicFiles.PUBLIC_LINK_SHARE_OWNER)).toBe(config.adminUsername)
+                expect(files[0].getProperty(oc.publicFiles.PUBLIC_LINK_SHARE_OWNER)).toBe(adminUsername)
```
individual-it, line 223: why do we need this change?

tests/publicLinkTest.js
```diff
@@ -624,7 +636,7 @@ describe('oc.publicFiles', function () {
               dProp
                 .appendElement('oc:public-link-item-type', '', 'folder')
                 .appendElement('oc:public-link-permission', '', permission)
-                .appendElement('oc:public-link-share-owner', '', config.adminUsername)
+                .appendElement('oc:public-link-share-owner', '', adminUsername)
```
individual-it, line 639: why changing that?

tests/helpers/sharingHelper.js
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ *
+ * @param {object} params
+ * @returns {string}
+ */
+const validateParams = function (params) {
+  let data = ''
+  if (params) {
+    if (params.path !== undefined && params.path !== null && params.path !== '') {
```
individual-it, line 95:
```suggestion
    if (params.path) {
```
I think that should do in JS see https://fellowtuts.com/jquery/check-undefined-null-empty-or-blank-variable-in-javascript/

tests/helpers/sharingHelper.js
```diff
@@ -0,0 +1,157 @@
...
+      data += `&path=${encodeURIComponent(params.path)}`
```
individual-it, line 96: using `URLsearchParams` might look nicer https://www.geeksforgeeks.org/node-js-urlsearchparams-api/

tests/helpers/sharingHelper.js
```diff
@@ -0,0 +1,157 @@
...
+const publicFilesEndPoint = '/remote.php/dav/public-files'
+
+/**
+ * returns the full and sanitized URL of the dav resource
+ *
+ * @returns {string}
+ */
+const getSharingEndPoint = function () {
+  return (process.env.PROVIDER_BASE_URL + shareEndPoint)
+    .replace(/([^:])\/{2,}/g, '$1/')
```
individual-it, line 17: same used in https://github.com/owncloud/owncloud-sdk/blob/bb47ee307afd7f0b112718fc08a48be6c0fbc805/tests/webdavHelper.js#L30

can we put it in some sanitise function used everywhere?

individual-it (changes requested):

tests/helpers/sharingHelper.js
```diff
@@ -0,0 +1,155 @@
...
+ * @param {object} data
+ * @returns {string}
+ */
+const validateParams = function (data) {
+  const params = new URLSearchParams()
+  if (data) {
+    if (data.path) {
+      params.append('path', data.path)
+    }
+    if (data.shareType !== undefined && data.shareType !== null && data.shareType !== '') {
```
individual-it, line 97:
```suggestion
    if (data.shareType) {
```

tests/helpers/sharingHelper.js
```diff
@@ -0,0 +1,155 @@
...
+    if (data.name) {
+      params.append('name', data.name)
+    }
+    if (data.password) {
+      params.append('password', data.password)
+    }
+    if (data.expireDate) {
+      params.append('expireDate', data.expireDate)
+    }
+    if (data.publicUpload !== undefined && data.publicUpload !== null && data.publicUpload !== '') {
```
individual-it, line 115:
```suggestion
    if (data.publicUpload) {
```

individual-it (changes requested):

tests/helpers/sharingHelper.js
saw-jan: might not work with `shareType` as share-type can be `0` which will be false

tests/helpers/sharingHelper.js
saw-jan: `publicUpload` can have value `false` so that would not work if we have something like `{ publicUpload: false }`.

individual-it (approved):

PR #743, opened 2021-02-22
use consumerVersionTags option instead of consumerVersionSelectors

individual-it:
pact-js 10.x (using pact V3) does not use (yet) `consumerVersionSelectors` but `consumerVersionTags` see https://github.com/pact-foundation/pact-js/blob/feat/v3.0.0/src/v3/verifier.ts

### DIFF
--- a/tests/providerTest.js
+++ b/tests/providerTest.js
@@ -47,12 +47,7 @@ describe('provider testing', () => {
       opts.publishVerificationResult = true
       opts.pactBrokerToken = process.env.PACTFLOW_TOKEN
       opts.enablePending = true
-      opts.consumerVersionSelectors = [
-        {
-          tag: process.env.DRONE_SOURCE_BRANCH,
-          latest: true
-        }
-      ]
+      opts.consumerVersionTags = process.env.DRONE_SOURCE_BRANCH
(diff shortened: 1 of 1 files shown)


saw-jan:
there is this code (may be this is useful)
```js
export interface ConsumerVersionSelector {
  pacticipant?: string
  tag?: string
  version?: string
  latest?: boolean
  all?: boolean
}
```

tests/providerTest.js
```diff
@@ -47,12 +47,7 @@ describe('provider testing', () => {
       opts.publishVerificationResult = true
       opts.pactBrokerToken = process.env.PACTFLOW_TOKEN
       opts.enablePending = true
-      opts.consumerVersionSelectors = [
-        {
-          tag: process.env.DRONE_SOURCE_BRANCH,
-          latest: true
-        }
-      ]
```
saw-jan, line 55: there is this code (may be this is useful)
```js
export interface ConsumerVersionSelector {
  pacticipant?: string
  tag?: string
  version?: string
  latest?: boolean
  all?: boolean
}
```

phil-davis (approved):

tests/providerTest.js
individual-it: it seems not to be used (yet)

Issue #723, opened 2021-02-10
harmonize request names


individual-it:
currently we have a lot of different names in `uponReceiving()`
we should make them more standard and look simmilar

PR #748, opened 2021-02-26
[tests-only] Consistent interaction names

saw-jan:
This PR makes pact interaction names more consistent
All interaction names follow the same pattern i.e. `.uponReceiving(a <METHOD> request ...)`

closes https://github.com/owncloud/owncloud-sdk/issues/723

### DIFF
--- a/tests/filesTest.js
+++ b/tests/filesTest.js
@@ -6,6 +6,7 @@ import { MatchersV3, XmlBuilder } from '@pact-foundation/pact/v3'
 
 describe('Main: Currently testing files management,', function () {
   const config = require('./config/config.json')
+  const username = config.adminUsername
 
   const {
     getContentsOfFileInteraction,
@@ -31,9 +32,9 @@ describe('Main: Currently testing files management,', function () {
   const { testFolder, testFile, testContent, nonExistentFile, nonExistentDir, owncloudURL } = config
   const testSubDir = testFolder + '/' + 'subdir'
 
-  const moveFileInteraction = function (provider, name, header, response) {
+  const moveFileInteraction = function (provider, requestName, header, response) {
     return provider
-      .uponReceiving('move existent file into same folder, ' + name)
+      .uponReceiving(`as '${username}', a MOVE request to move existent file into same folder, ${requestName}`)
       .withRequest({
         method: 'MOVE',
         path: webdavPath(`${testFolder}/${encodeURI('中文.txt')}`),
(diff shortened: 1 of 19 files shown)


tests/userTest.js
```diff
@@ -549,7 +549,7 @@ describe('Main: Currently testing user management,', function () {
     await getCurrentUserInformationInteraction(provider)
     await getUsersInteraction(
       provider,
-      'search for a user that exists',
+      'check user existence with existing user',
```
Talank, line 552:
```suggestion
      'check user existence with existing users',
```

tests/userTest.js
```diff
@@ -575,7 +575,7 @@ describe('Main: Currently testing user management,', function () {
     await getCurrentUserInformationInteraction(provider)
     await getUsersInteraction(
       provider,
-      'search for a user that doesn\'t exists',
+      'check user existence with non-existing user',
```
Talank, line 578:
```suggestion
      'check user existence with non-existing users',
```

tests/userTest.js
saw-jan: `users` would be incorrect as it is more like checking user existence using one non-existing user in the request

tests/userTest.js
saw-jan: `users` would be incorrect as it is more like checking user existence using one non-existing user in the request

Talank:
LGTM

dpakach (commented):
lgtm :+1:

tests/filesTest.js
```diff
@@ -1337,12 +1336,12 @@ describe('Main: Currently testing files management,', function () {
...
         })
       await provider
-        .uponReceiving('PROPFIND file info, fileId')
+        .uponReceiving('a PROPFIND request to file info, fileId')
         .withRequest({
           method: 'PROPFIND',
           path: webdavPath(`${testFolder}/${testFile}`),
           headers: validAuthHeaders,
-          body: new XmlBuilder('1.0', '', 'd:profind').build(dPropfind => {
+          body: new XmlBuilder('1.0', '', 'd:propfind').build(dPropfind => {
```
individual-it, line 1344: good catch, but how did it work before?

tests/pactHelper.js
```diff
@@ -342,7 +342,7 @@ const getUserInformationAsAdminInteraction = function (provider) {
 }
 
 const getCapabilitiesWithInvalidAuthInteraction = function (provider, username = config.adminUsername, password = config.invalidPassword) {
-  return provider.uponReceiving(`a capabilities GET request with invalid authentication by ${username} with ${password}`)
+  return provider.uponReceiving(`a GET request to get capabilities with invalid auth by '${username}' with '${password}'`)
```
individual-it, line 345: should be `as user ..` if we follow the same structure like in the other requests

individual-it (changes requested):
just one small change request
and for discussion: should we put `as ...` into every interaction name?

tests/filesTest.js
saw-jan: this test is skipped

tests/pactHelper.js
saw-jan: the decision was made to use `as <username>, a <METHOD> request ...`

individual-it (approved):

PR #754, opened 2021-03-03
[tests-only] Do not set skeleton directory in CI

saw-jan:
This PR removes line/step to set skeleton directory in owncloud server config.
Skeleton files are not needed for the tests and preventing skeleton files from being created for each user creation will make the test run-time faster (testing against owncloud10).

Also removes other unnecessary configs

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -84,21 +84,14 @@ def installCore(version):
 
     return [stepDefinition]
 
-def setupServerAndApp(logLevel):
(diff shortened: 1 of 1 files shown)


dpakach (approved):

dpakach (commented):
While you're at it can you also remove other unnecessary configs

individual-it:
Well spotted :+1:

PR #578, opened 2020-09-20
Add enable/disable user api

rpocklin:
I have added `enable` and `disable` user API endpoints, these were missing and needed for some functionality I am currently working on.  I have written the tests without regard to your pact.io work in `tests/userTest.js` which tests both methods in succession.

PR #759, opened 2021-03-04
add enable/disable user api (no-fork)

individual-it:
same as #578 but branch-name has not `/` because that does not work in CI currently


kulmann (approved):
LGTM 👍

Issue #747, opened 2021-02-24
adjust pact interactions so that they also pass provider side (ocis)


individual-it:
follow up of #740
sub task of #644

Currently we run all tests of the consumer pact against the ocis provider but most of them fail, see #740 because the contract from the consumer side is not correct or missing given steps.
All failing interactions need to be adjusted, so that the provider accepts them.

saw-jan:
Tests to check

- [ ] unauthorized
  - [ ] appTest.js
  - [ ] capabilitiesTest.js
  - [ ] filesTest.js
  - [ ] groupTest.js
  - [ ] sharingTest.js
  - [ ] userTest.js

- [ ] appTest.js
 
 - [ ] capabilitiesTest.js

 - [ ] filesTest.js

 - [ ] filesTrashTest.js

 - [ ] fileVersionTest.js

 - [ ] groupTest.js

 - [ ] loginTest.js

 - [ ] pactHelper.js

 - [ ] publickLinkTest.js

 - [ ] requestsOcsTest.js
 
 - [ ] shareRecipientTest.js

 - [ ] sharingTest.js

 - [ ] sharingWithAttributesTest.js

 - [ ] signedUrlIntegrationTest.js

 - [ ] userTest.js

PR #760, opened 2021-03-04
[tests-only] Adjust groupTest interactions for oCIS server

saw-jan:
part of https://github.com/owncloud/owncloud-sdk/issues/644 and https://github.com/owncloud/owncloud-sdk/issues/747

adjust interactions for oCIS:
- groupTest.js

### DIFF
--- a/tests/groupTest.js
+++ b/tests/groupTest.js
@@ -17,6 +17,7 @@ describe('Main: Currently testing group management,', function () {
 
   async function getGroupsInteraction (provider) {
     await provider
+      .given('group exists', { groupName: 'admin' })
       .given('group exists', { groupName: config.testGroup })
       .uponReceiving(`as '${username}', a GET request to get all groups`)
       .withRequest({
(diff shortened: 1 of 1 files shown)


Talank (commented):
LGTM

tests/groupTest.js
```diff
@@ -138,28 +141,41 @@ describe('Main: Currently testing group management,', function () {
...
   })
 
   it('checking method : deleteGroup with a non-existent group', async function () {
-    await deleteGroupInteraction(provider, config.nonExistentGroup)
+    await deleteGroupInteraction(
+      provider,
+      config.nonExistentGroup,
+      meta => ocsMeta(
+        meta,
+        MatchersV3.regex('(failure|error)', 'failure'),
```
individual-it, line 174: is this a difference in ocis vs. oc10? is there an issue open for that?

tests/groupTest.js
```diff
@@ -32,19 +32,22 @@ describe('Main: Currently testing group management,', function () {
         headers: xmlResponseHeaders,
         body: new XmlBuilder('1.0', '', 'ocs').build(ocs => {
           ocs.appendElement('meta', '', (meta) => {
-            return ocsMeta(meta, 'ok', '100')
+            return ocsMeta(meta, 'ok', '100', MatchersV3.regex('(OK)?', ''))
```
individual-it, line 35: does the client need this response, if not we should ignore it altogether and never send it back

tests/groupTest.js
```diff
@@ -88,7 +89,6 @@ describe('Main: Currently testing group management,', function () {
       await oc.login()
       return oc.groups.getGroups().then(data => {
         expect(typeof (data)).toBe('object')
-        expect(data.indexOf('admin')).toBeGreaterThan(-1)
```
individual-it, line 91: It does not feel good for me to remove this check
maybe we can add the "admin" group in ocis

individual-it (changes requested):

tests/groupTest.js
saw-jan: It seems that client needs `meta.message` too. If we are to append `message` element only if needed like this
```js
if (Message !== null) {
    meta.appendElement('message', '', Message)
}
```
then the consumer tests will fail with following error
```js
TypeError: Cannot convert undefined or null to object
        at Function.keys (<anonymous>)

      392 |       ret = meta.message
      393 | 
    > 394 |       if (Object.keys(meta.message).length === 0) {
          |                  ^
      395 |         // no error message returned, return the whole message
      396 |         ret = json
      397 |       }
```
At least empty `message` element needs to be appended

individual-it (approved):

Issue #722, opened 2021-02-10
review returns of state handlers


individual-it:
do we need `return Promise.resolve({ description: 'file created' })` in the state handlers? maybe we can just delete all the lines

PR #762, opened 2021-03-04
[Tests-Only] Remove use of promise from sync functions

dpakach:
Fixes https://github.com/owncloud/owncloud-sdk/issues/722
Remove use of Promise from `provider.given()`.
the documentation is not very clear about what is clear format, but since the `given()` function doesn't support promises, we need to remove the Promise.resolve() from return statmenets.

### DIFF
--- a/tests/providerTest.js
+++ b/tests/providerTest.js
@@ -36,6 +36,8 @@ describe('provider testing', () => {
 
   chai.use(chaiAsPromised)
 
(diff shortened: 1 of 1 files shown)


saw-jan (approved):
LGTM :+1:

swoichha (commented):
lgtm

individual-it:
:+1: fixes #722

PR #769, opened 2021-03-11
[tests-only] Adjust fileTest  interaction for oCIS

SwikritiT:
part of #644 and #747

adjust interactions for oCIS:

* fileTest

### DIFF
--- a/tests/filesTest.js
+++ b/tests/filesTest.js
@@ -635,8 +635,7 @@ describe('Main: Currently testing files management,', function () {
           }
         })
         .willRespondWith({
-          status: 201,
-          headers: htmlResponseHeaders
+          status: 201
         })
       return provider.executeTest(async () => {
         const oc = createOwncloud(config.testUser, config.testUserPassword)
@@ -976,8 +975,7 @@ describe('Main: Currently testing files management,', function () {
           )
         },
         {
-          status: 201,
-          headers: htmlResponseHeaders
+          status: 201
         })
 
       return provider.executeTest(async () => {
@@ -1027,8 +1025,7 @@ describe('Main: Currently testing files management,', function () {
           }
         })
         .willRespondWith({
-          status: 201,
-          headers: htmlResponseHeaders
+          status: 201
         })
 
       return provider.executeTest(async () => {
@@ -1082,8 +1079,7 @@ describe('Main: Currently testing files management,', function () {
           }
         })
         .willRespondWith({
-          status: 201,
-          headers: htmlResponseHeaders
+          status: 201
         })
 
       return provider.executeTest(async () => {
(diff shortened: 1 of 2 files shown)


Talank (commented):
the changes LGTM, but there seem to be some conflicts so please rebase and resolve it.

tests/filesTest.js
```diff
@@ -629,7 +631,7 @@ describe('Main: Currently testing files management,', function () {
           headers: {
             authorization: getAuthHeaders(config.testUser, config.testUserPassword),
             Destination: MatchersV3.fromProviderState(
-              `\${providerBaseURL}${destinationWebDavPath}`,
+              `\${providerBaseURL}/${destinationWebDavPath}`,
```
saw-jan, line 634: this change may not be required as it depends on `PROVIDER_BASE_URL` env variable.
you can have trailing`/` as `PROVIDER_BASE_URL=https/localhost:9200/` while running provider tests

tests/filesTest.js
```diff
@@ -1082,8 +1085,7 @@ describe('Main: Currently testing files management,', function () {
           }
         })
         .willRespondWith({
-          status: 201,
-          headers: htmlResponseHeaders
```
saw-jan, line 1086: NOTE: a COPY request's reponse doesn't have content-type header in ocis server as oc10 does.
As client doesn't need `content-type` here, I feel safe to remove this line.

or any thoughts?

SwikritiT: Yes , I removed the `headers` part , only `status` is present.

swoichha (commented):
LGTM

individual-it (approved):

tests/filesTest.js
```diff
@@ -601,6 +601,8 @@ describe('Main: Currently testing files management,', function () {
       })
     })
 
+    // [OCIS] ocis returns `text/plain` content-type and not `text/html` like oc10 while sending move request
+    // https://github.com/owncloud/ocis/issues/1793
```
saw-jan, line 605: as `MOVE` request doesn't send any payload (response body), it will also be safe to remove `headers: htmlResponseHeaders` from here https://github.com/owncloud/owncloud-sdk/blob/e3237cbbadf8a61369d0a219572eadeda907a535/tests/filesTest.js#L639-L642

tests/filesTest.js
```diff
@@ -943,6 +945,8 @@ describe('Main: Currently testing files management,', function () {
     })
   })
 
+  // [OCIS] ocis returns `text/plain` content-type and not `text/html` like oc10 while sending move request
+  // https://github.com/owncloud/ocis/issues/1793
```
saw-jan, line 949: as `MOVE` request doesn't send any payload (response body), it will also be safe to remove `headers: htmlResponseHeaders` from here https://github.com/owncloud/owncloud-sdk/blob/e3237cbbadf8a61369d0a219572eadeda907a535/tests/filesTest.js#L982-L985

saw-jan (changes requested):

saw-jan (approved):

Issue #777, opened 2021-03-15
[tests-only] Make every request urls to have `/` at the beginning


saw-jan:
In .drone.start server urls are `http://owncloud/` for oc10 and `https://ocis:9200` for ocis.
Some request urls don't have `/` at the beginning which are making those tests fail with `400` bad request in ocis server.
So, the following fixes are needed:
 - change provider server urls to not have trailing `/` in .drone.star
 - make sure every requests in tests have `/` at the beginning
 - make use of function to get provider's base url

PR #778, opened 2021-03-15
[tests-only] Fix trailing '/' of the provider server url and check '/' at the beginning of requests in tests

saw-jan:
This PR makes sure that every request endpoints in the tests have `/` at the beginning
and the server base url do not have trailing `/`
And also makes sure that while testing against `oCIS` server there is no any `404` bad request errors

closes https://github.com/owncloud/owncloud-sdk/issues/777

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -51,7 +51,7 @@ def buildSystem():
             ],
     }]
 
-def prepareTestConfig(subFolderPath = '/'):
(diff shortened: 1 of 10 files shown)


kiranparajuli589:
lgtm :+1:

swoichha (commented):
LGTM

phil-davis (approved):

PR #783, opened 2021-03-17
split provider tests into multiple pacts to be able to verify some interactions

individual-it:
on the provider side we have different interactions
1. interactions that work on both oc10 & ocis
2. interactions that are "pending" on oc10 but should work on ocis
3. interactions that are "pending" on ocis but should work on oc10
4. interactions that are "pending" on both ocis & oc10

we don't want CI to fail for the interactions that are "pending" but we want CI to fail if an interaction that was already verified starts to fail. Pact.io has an system for that called [pending pacts](https://docs.pact.io/pact_broker/advanced_topics/pending_pacts/) but it seems to work only on pact level and not on interactions level

So in this PR I create 4 different pacts for the different buckets of interactions
when running the provider tests pacts that are allowed to fail are marked as "pending" others not
when creating a test `createProvider` need to be called with two parameters, depending on which provider-version the test is still pending

issues with this approach:
- when a test contains of multiple interactions on the consumer side (basically all tests) and one of the interactions has a problem (e.g. get capabilities on OCIS) all interactions of that test will fall in the "pending" bucket
- its a bit tedious to enable more test for a provider version


jest.setup.js
```diff
@@ -0,0 +1,20 @@
...
+  provider = createProvider(true, true)
+  await provider.executeTest(async () => {
+  })
+  provider = createProvider(true, false)
+  await provider.executeTest(async () => {
+  })
+  provider = createProvider(false, true)
+  await provider.executeTest(async () => {
+  })
+}
```
dpakach, line 20: Do we need this? there are no tests here

individual-it: This is to create empty pact files.
That way we make sure there is a file for every combination of verified/pending interactions
I've wrote a comment in the file

saw-jan (commented):
FGTM

dpakach (approved):
lgtm :+1:

jasson99 (commented):
LGTM

Issue #788, opened 2021-03-17
split public link create test from list content test 


individual-it:
`'should list the folder contents'` and `'should download files from a public shared folder'` also test `shareFileWithLink()` see https://github.com/owncloud/owncloud-sdk/blob/c024777722794bd004adfb4e368e15004adfe38b/tests/publicLinkTest.js#L227
but those tests should only test what they claim to test, list folder and download a file
`shareFileWithLink()` should be tested in a separate test and the list & download tests should not rely on `shareFileWithLink()` and the corresponding interactions

Issue #780, opened 2021-03-16
[Tests-Only] Fix pact interactions for provider side tests


dpakach:
Part of https://github.com/owncloud/owncloud-sdk/issues/644
#### For oc10 backend
- [x] - tests/FilesTrashTest.js https://github.com/owncloud/owncloud-sdk/pull/798 
- [x] - tests/FilesVersionTest.js https://github.com/owncloud/owncloud-sdk/pull/798 
- [x] - tests/PublicLinkTest.js https://github.com/owncloud/owncloud-sdk/pull/791
- [x] - tests/FilesTest.js  https://github.com/owncloud/owncloud-sdk/pull/797
- [x] - tests/unauthorized/FilesTest.js https://github.com/owncloud/owncloud-sdk/pull/782

#### For ocis backend
- [x] - tests/FilesTrashTest.js https://github.com/owncloud/owncloud-sdk/pull/798 
- [x] - tests/FilesVersionTest.js https://github.com/owncloud/owncloud-sdk/pull/798 
- [x] - tests/PublicLinkTest.js https://github.com/owncloud/owncloud-sdk/pull/815
- [x] - tests/FilesTest.js https://github.com/owncloud/owncloud-sdk/pull/811
- [x] - tests/unauthorized/FilesTest.js https://github.com/owncloud/owncloud-sdk/pull/815

saw-jan:
All above-listed tests' interactions have been adjusted for servers

PR #791, opened 2021-03-18
[tests-only] Adjust unskipped publicLink tests

saw-jan:
part of https://github.com/owncloud/owncloud-sdk/issues/644 and https://github.com/owncloud/owncloud-sdk/issues/780

Adjust pact interaction for unskipped tests in publicLinkTest.js

Fixes https://github.com/owncloud/owncloud-sdk/issues/788

### DIFF
--- a/tests/helpers/providerStateHelper.js
+++ b/tests/helpers/providerStateHelper.js
@@ -246,6 +246,31 @@ const givenResourceIsShared = async (
   }
 }
 
+/**
+ * provider state: creates a public link share
+ *
+ * @param {object} provider
+ * @param {string} username
+ * @param {sring} userPassword
+ * @param {sring} path
+ * @param {object} optionalParams
+ */
+const givenPublicShareExists = (
+  provider,
+  username,
+  userPassword,
+  path,
+  optionalParams
+) => {
+  return provider
+    .given('public link share is created', {
+      username,
+      userPassword,
+      path,
+      ...optionalParams
+    })
+}
+
 module.exports = {
   givenUserExists,
   givenGroupExists,
@@ -255,5 +280,6 @@ module.exports = {
   givenGroupShareExists,
   givenPublicShareExists,
   givenFileFolderIsCreated,
-  givenResourceIsShared
+  givenResourceIsShared,
+  givenPublicShareExists
 }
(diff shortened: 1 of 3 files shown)


tests/helpers/providerStateHelper.js
```diff
@@ -33,7 +33,7 @@ const givenGroupExists = (provider, group) => {
  */
 const givenFileExists = (provider, username, password, resource) => {
   return provider
-    .given('file exists', { username, password, fileName: encodeURIComponent(resource) })
+    .given('file exists', { username, password, fileName: encodeURIComponent(resource).split('%2F').join('/') })
```
dpakach, line 36: Why is this required here?, if it doesn't likes encoded filename then we can use `decodeURIComponent`

tests/providerTest.js
```diff
@@ -314,6 +314,29 @@ describe('provider testing', () => {
             date: date
           }
         }
+      },
+      'public link share is created': (setup, parameters) => {
+        if (setup) {
+          const { username, userPassword, path, shareType, permissions, password } = parameters
```
dpakach, line 320: share type for the public link will always be the same you don't have to pass it from parameters

tests/helpers/providerStateHelper.js
saw-jan: It needs to encode files but there can be a path like `folder/file.txt`. In that case, it will also encode `/` which we don't want happen.

dpakach (commented):
lgtm :+1:

Talank (commented):
LGTM

individual-it (changes requested):
for those tests that pass on the provider side on oc10 please set the first attribute of `createProvider` to false that will put them into the `verified` bucket and fail CI if those tests start to fail later
I don't expect any provider tests pass completely on ocis because the capabilities still return the wrong content type

tests/helpers/providerStateHelper.js
individual-it, line 36: use it in a function like in https://github.com/owncloud/owncloud-sdk/pull/801/files#diff-a16db6b5aac8f41269353f9be2487f36a04fd3e16286cd2c0edf4ce46a489aa5R24

individual-it (changes requested):

individual-it (approved):

PR #810, opened 2021-04-20
switch to official redis

wkloucek:
## Description
switch to official Redis docker image in CI

## Related Issue

- None

## Motivation and Context

Use only maintained docker images


individual-it (commented):
we will not need any redis after #835

wkloucek:
> we will not need any redis after #835
even better :+1:

PR #811, opened 2021-04-22
[tests-only] Adjust pact interactions for oCIS server

saw-jan:
part of https://github.com/owncloud/owncloud-sdk/issues/644 and https://github.com/owncloud/owncloud-sdk/issues/780

adjust interactions for oCIS:
- filesTest

### DIFF
--- a/tests/filesTest.js
+++ b/tests/filesTest.js
@@ -83,9 +83,12 @@ describe('Main: Currently testing files management,', function () {
                     dProp.appendElement('d:resourcetype', '', dResourcetype => {
                       dResourcetype.appendElement('d:collection', '', '')
                     })
-                      .appendElement('d:quota-used-bytes', '', MatchersV3.number(55))
-                      .appendElement('d:quota-available-bytes', '', MatchersV3.number(3))
-                      .appendElement('d:getetag', '', MatchersV3.regex('"[a-z0-9]+"', '"5f8d0ce8c62b5"'))
+                      .appendElement('d:getetag', '',
+                        MatchersV3.regex(
+                          '"[a-z0-9]+"',
+                          '"5f8d0ce8c62b5"'
+                        )
+                      )
                   })
                     .appendElement('d:status', '', 'HTTP/1.1 200 OK')
                 })
@@ -140,7 +143,7 @@ describe('Main: Currently testing files management,', function () {
           .appendElement(
             'd:href', '',
             MatchersV3.regex(
-              `*.\\/remote\\.php\\/webdav\\/${testFolder}\\/${item.name}`,
+              `.*\\/remote\\.php\\/webdav\\/${testFolder}\\/${item.name}(\\/)?`,
               `/remote.php/webdav/${testFolder}/${item.name}`
             )
           )
@@ -151,23 +154,23 @@ describe('Main: Currently testing files management,', function () {
                   'd:getlastmodified', '',
                   MatchersV3.date('EEE, d MMM yyyy HH:mm:ss z', 'Mon, 19 Oct 2020 03:50:00 GMT')
                 )
-                .appendElement('d:getcontentlength', '', MatchersV3.number(11))
                 .appendElement('d:resourcetype', '', resourceType)
                 .appendElement(
                   'd:getetag', '',
-                  MatchersV3.regex('"[a-z0-9]+"', '"3986cd55c130a4d50ff0904bf64aa27d"')
+                  MatchersV3.regex(
+                    '"[a-z0-9]+"',
+                    '"3986cd55c130a4d50ff0904bf64aa27d"'
+                  )
                 )
-                .appendElement('d:getcontenttype', '', 'text/plain')
+              if (item.type === 'file') {
+                dProp
+                  .appendElement('d:getcontentlength', '', MatchersV3.number(11))
+                  .appendElement('d:getcontenttype', '',
+                    MatchersV3.regex('text\\/plain(; charset=utf-8)?', 'text/plain')
+                  )
+              }
             })
               .appendElement('d:status', '', 'HTTP/1.1 200 OK')
-          })
-          .appendElement('d:propstat', '', dPropstat => {
-            dPropstat.appendElement('d:prop', '', dProp => {
-              dProp
-                .appendElement('d:quota-used-bytes', '', '')
-                .appendElement('d:quota-available-bytes', '', '')
-            })
-              .appendElement('d:status', '', 'HTTP/1.1 404 Not Found')
           })
       })
     }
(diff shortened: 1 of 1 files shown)


swoichha (commented):
lgtm

tests/filesTest.js
```diff
@@ -83,9 +84,7 @@ describe('Main: Currently testing files management,', function () {
                     dProp.appendElement('d:resourcetype', '', dResourcetype => {
                       dResourcetype.appendElement('d:collection', '', '')
                     })
-                      .appendElement('d:quota-used-bytes', '', MatchersV3.number(55))
-                      .appendElement('d:quota-available-bytes', '', MatchersV3.number(3))
-                      .appendElement('d:getetag', '', MatchersV3.regex('"[a-z0-9]+"', '"5f8d0ce8c62b5"'))
+                      .appendElement('d:getetag', '', MatchersV3.string('&quot;5f8d0ce8c62b5&quot;'))
```
individual-it, line 87: I would suggest to leave regex for etags, to make sure the etags are valid, meaning are quoted `"`

tests/filesTest.js
```diff
@@ -140,7 +139,7 @@ describe('Main: Currently testing files management,', function () {
           .appendElement(
             'd:href', '',
             MatchersV3.regex(
-              `*.\\/remote\\.php\\/webdav\\/${testFolder}\\/${item.name}`,
+              `.*\\/remote\\.php\\/webdav\\/${testFolder}\\/${item.name}(\\/)?`,
```
individual-it, line 142: good catch `*.` does not make sense

tests/filesTest.js
```diff
@@ -151,24 +150,21 @@ describe('Main: Currently testing files management,', function () {
...
-                  MatchersV3.regex('"[a-z0-9]+"', '"3986cd55c130a4d50ff0904bf64aa27d"')
+                  MatchersV3.string('&quot;3986cd55c130a4d50ff0904bf64aa27d&quot;')
                 )
-                .appendElement('d:getcontenttype', '', 'text/plain')
+              if (item.type === 'file') {
+                dProp
+                  .appendElement('d:getcontentlength', '', MatchersV3.number(11))
+                  .appendElement('d:getcontenttype', '',
+                    MatchersV3.regex('text\\/plain(; charset=utf-8)?', 'text/plain')
+                  )
```
individual-it, line 163: no contenttype and length for folders? is that the same in oc10? if not we need to open a bug report

tests/filesTest.js
```diff
@@ -983,19 +995,41 @@ describe('Main: Currently testing files management,', function () {
...
+      await givenUserExists(provider, config.testUser, config.testUserPassword)
+      const reqPath = path
+      let examplePath, regexPath
+      if (path !== '/') {
+        examplePath = '/' + path + '/'
+        regexPath = examplePath.replace(/\//g, '\\\\/')
+        await givenFolderExists(provider, config.testUser, config.testUserPassword, reqPath)
+      } else {
+        examplePath = path
+        regexPath = examplePath.replace(/\//g, '\\\\/')
```
individual-it, line 1013: can be moved outside of the if condition

tests/helpers/providerStateHelper.js
```diff
@@ -286,6 +286,19 @@ const givenTagIsAssignedToFile = (provider, username, password, fileName, tagNam
...
 
+/**
+ * uploads file using TUS protocol
+ *
+ * @param {object} provider
+ * @param {string} username
+ * @param {sring} password
+ * @param {sring} fileName
+ */
+const givenFileUploadedWithTUS = (provider, username, password, fileName) => {
```
individual-it, line 297: why do we need that a given step? if its a given action then we don't care how the file is uploaded, if we want to test the TUS upload then it should not be a `given` step

tests/providerTest.js
```diff
@@ -403,6 +403,29 @@ describe('provider testing', () => {
           chai.assert.fail('Failed to assign last created tag to last created file')
         }
       }
+    },
+    'a file is uploaded using TUS protocol': (setup, parameters) => {
```
individual-it, line 407: you only the creation of a TUS file, not an actual upload 
CC @swoichha

individual-it (changes requested):
it might have been good to split this PR into multiple

tests/filesTest.js
saw-jan: Yeah, same for both oC10 and oCIS

tests/providerTest.js
saw-jan: @individual-it @swoichha 
The tests are like `returns TUS support information when TUS headers are set for a list call`
Please, guide me what needs to be done as given for such tests.

kiranparajuli589:
lgtm 👍🏼

tests/helpers/providerStateHelper.js
saw-jan: @individual-it the given test is not for TUS upload, it is to list directory and check TUS support information.

tests/filesTest.js
```diff
@@ -151,24 +154,24 @@ describe('Main: Currently testing files management,', function () {
...
                   MatchersV3.date('EEE, d MMM yyyy HH:mm:ss z', 'Mon, 19 Oct 2020 03:50:00 GMT')
                 )
-                .appendElement('d:getcontentlength', '', MatchersV3.number(11))
                 .appendElement('d:resourcetype', '', resourceType)
                 .appendElement(
                   'd:getetag', '',
-                  MatchersV3.regex('"[a-z0-9]+"', '"3986cd55c130a4d50ff0904bf64aa27d"')
+                  MatchersV3.regex(
+                    '"[a-z0-9]+"',
+                    '&quot;3986cd55c130a4d50ff0904bf64aa27d&quot;'
```
individual-it, line 162: why do we need to use `&quot;` here and not simply `"`?

tests/filesTest.js
```diff
@@ -627,6 +640,7 @@ describe('Main: Currently testing files management,', function () {
       })
     })
 
+    // [oCIS] 500 server error
```
individual-it, line 643: is there an issue open for that?

individual-it (changes requested):

tests/filesTest.js
saw-jan: was just a copy paste. I will update it

tests/filesTest.js
saw-jan: that should be by mistake.

individual-it (approved):

individual-it:
I've squashed the commits, this PR is ready to be merged when CI is green

Issue #834, opened 2021-05-06
chage ocis storage driver to "ocis"


individual-it:
- [ ] change storage driver
- [ ] revert the deletion part of https://github.com/owncloud/owncloud-sdk/pull/755

PR #755, opened 2021-03-03
[tests-only] Delete user data directory when testing against oCIS server

saw-jan:
In oCIS, deleting users via API doesn't delete the user's data directory.
When recreating deleted user, previously added files/folders are also restored, which causes unexpected failures in oCIS provider tests.

This PR aims to cleanup the user's data directory every time before the user is recreated.

POC https://drone.owncloud.com/owncloud/owncloud-sdk/2275/2/8 (L:73, L:115)

Closes https://github.com/owncloud/owncloud-sdk/issues/814

PR #835, opened 2021-05-06
[tests-only] Change ocis storage driver from owncloud to ocis

saw-jan:
changes:
- change storage driver to `ocis`
- revert changes made in https://github.com/owncloud/owncloud-sdk/pull/755
- bump ocis commit-id

Closes https://github.com/owncloud/owncloud-sdk/issues/834

### DIFF
--- a/tests/providerTest.js
+++ b/tests/providerTest.js
@@ -207,12 +207,18 @@ describe('provider testing', () => {
         chai.assert.strictEqual(
           result.status, 200, `creating user '${parameters.username}' failed`
         )
-        // a hack for https://github.com/owncloud/ocis/issues/1675
-        fetch(providerBaseUrl + '/ocs/v2.php/cloud/capabilities',
-          {
-            method: 'GET',
-            headers: { authorization: getAuthHeaders(parameters.username, parameters.password) }
-          })
+        // a hack for
+        // https://github.com/owncloud/ocis/issues/1675
+        // https://github.com/owncloud/ocis/issues/2027
+        if (isRunningWithOCIS()) {
+          for (let count = 1; count <= 2; count++) {
+            fetch(providerBaseUrl + '/ocs/v2.php/cloud/capabilities',
+              {
+                method: 'GET',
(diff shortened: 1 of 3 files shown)


individual-it:
redis also can be deleted

tests/providerTest.js
```diff
@@ -223,9 +204,13 @@ describe('provider testing', () => {
...
-        chai.assert.strictEqual(
-          result.status, 200, `creating user '${parameters.username}' failed`
-        )
+
+        const { message } = getOCSMeta(result)
+        if (result.status !== 200 && message !== 'account already exists') {
+          chai.assert.strictEqual(
+            result.status, 200, `creating user '${parameters.username}' failed`
+          )
+        }
```
saw-jan, line 213: recreating a user should work but sometimes it fails with `account already exists` mostly with share user `testABC`.
So have to add this code block.

But will also work with this delete code if it is not to removed:
https://github.com/owncloud/owncloud-sdk/blob/8623274eed976753b03a6d31572d44b29b18b5a4/tests/providerTest.js#L213-L215

saw-jan: Related issue in ocis: https://github.com/owncloud/ocis/issues/2027

individual-it: the user should have done at least a request to `/ocs/v2.php/cloud/capabilities` , see the hack below. Maybe that is not enough.
Can we change the hack below an make it a webdav `PROPFIND` request? will that create the user correctly or do we need to create a file / folder?

saw-jan: @individual-it We need to make any requests twice and then only the user will be deleted successfully. (requesting `capabilities` twice will work)
Should we do that?

individual-it: yes, let's try that and please mention that also in owncloud/ocis#2027

individual-it (approved):

phil-davis (approved):

tests/providerTest.js
phil-davis: Comment added to the issue: https://github.com/owncloud/ocis/issues/2027#issuecomment-834116526

PR #840, opened 2021-05-10
Update axios version to 0.21.1

dpakach:
update axios version to 0.21.1

Fix related issues with the new version
- After updating the webdav client in https://github.com/owncloud/owncloud-sdk/pull/735 (which also uses Axios in the backend), the request interceptor which was already in place (to add the authtoken in every request) also kicks in for the dav requests.
 While this is ok for regular file operations, in public link, this creates a problem since public link uses different form of auth ( eg. public:password). That was fixed in this PR

- The auth headers were added in two different forms (`authorization` and `Authorization`). Because of this, after the request interceptor adds new header the actual header sent were in the form `Authorization: Bearer <token>, Bearer <token>`
  This pr changes all auth headers to `Authorization`

- Remove `Authorization: undefined` from the interceptor as axios throws error when it's set.

### DIFF
--- a/src/helperFunctions.js
+++ b/src/helperFunctions.js
@@ -52,6 +52,14 @@ class helpers {
     return this.instance
   }
 
+  noAuth () {
+    axios.interceptors.request.use(config => {
+      config.extraReqParams = config.extraReqParams || {}
+      config.extraReqParams.dontUseDefaultAuth = true
+      return config
+    })
+  }
+
   /**
    * sets the username
    * @param   {string | null}    authHeader    authorization header; either basic or bearer or what ever
@@ -59,6 +67,10 @@ class helpers {
   setAuthorization (authHeader) {
     this._authHeader = authHeader
     axios.interceptors.request.use(config => {
+      if (config.extraReqParams && config.extraReqParams.dontUseDefaultAuth) {
+        delete config.extraReqParams.dontUseDefaultAuth
+        return config
+      }
       if (authHeader && authHeader.startsWith('Bearer ')) {
         config.headers.Authorization = authHeader
       }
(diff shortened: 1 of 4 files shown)


src/helperFunctions.js
```diff
@@ -61,8 +61,6 @@ class helpers {
     axios.interceptors.request.use(config => {
       if (authHeader && authHeader.startsWith('Bearer ')) {
         config.headers.Authorization = authHeader
-      } else {
-        config.headers.Authorization = undefined
```
individual-it, line 65: why  this change?

dpakach:
All tests pass in web (with oc10) using this version of owncloud sdk. 
https://drone.owncloud.com/owncloud/web/15375

The unit tests are failing (that might require changes on web)
Some tests fail on ocis (in drone but pass locally). That needs some investigation.

src/publicLinkAccess.js
```diff
@@ -38,8 +38,11 @@ class PublicFiles {
     const headers = this.helpers.buildHeaders(false)
     const url = this.getFileSharePath(tokenAndPath)
 
+    // Dont use default bearer token for public links in web
+    // this will lead to the link not being accessible
+    this.helpers.noAuth()
```
LukasHirt, line 43: Is this enough if it's called only here? Don't we need to call it in all the other public methods?

dpakach: @LukasHirt, From what I can tell, when you try to access a public link from a WebUI in web, it first tries to do a PROPFIND without any auth. if it gets 401 (If public link is password protected), it will give the enter password dialog. After that it uses basic auth for all other requests. So if we just handle that for the first PROPFIND request it is not really necessary for other requests.

LukasHirt: I see. Interesting...

PR #853, opened 2021-05-28
properly catch JSON parse errors in _makeOCSrequest

rpocklin:
When receiving invalid JSON (or non-JSON datatype) responses from a call to **_makeOCSrequest** in this library, it can break the Javascript runtime flow and end the current operation being executed, unable to catch and deal with the exception.

This is because there is line `JSON.parse(body)`, which is in a catch block, can also throw an exception.

Once this is merged and released and you update the version of this being used in **owncloud/web** you can then remove this complicated workaround:
https://github.com/owncloud/web/blob/d0a3248aa802e8cd4ad33e9b871dd69436d4ea41/rollup.config.js#L57-L59


kulmann (approved):
Good catch 👍

Issue #854, opened 2021-05-28
Global interceptor issue when using owncloud-sdk


rpocklin:
Currently once you instantiate `new OwnCloud()` it will call the `init` method automatically, which then makes a call to `setAuthorization` which creates a global axios interceptor (see https://github.com/owncloud/owncloud-sdk/blob/b9c34546fdfba150b68d2d1352033811ec48b11b/src/owncloud.js#L37-L57).

This is undesirable as it will affect any other requests in the Javascript runtime using axios and even once a call to `logout` is made there this interceptor is not unbound (aka ejected) to restore it to the original state.

This creates havoc when attempting to use more than one instance of Owncloud to manage multiple users sessions at different times in a single page application.

I have raised a PR here to fix this issue and have tested this locally and it works as expected.

rpocklin:
I can see this has been merged, but am not able to easily track the released versions of `owncloud-sdk` - when I look at *owncloud/web* it currently references "1.0.0-2296" but cannot see a branch or tag in this repo that corresponds - would be great to have your releases tagged for external visibility.

kulmann:
Fully agreed, deserves tagging. At the moment releases in this repository happen automatically by merging to master. The release on npmjs.com will have the version `1.0.0-<some-automatically-increasing-number>`. Turned out just recently (last week) that this is horrible for reverting to another version of the sdk. For now you can only have a look at the `Versions` page on npmjs.com and guesstimate if the version 2296 (which says `2 months ago`) does not include your PR, yet.

kulmann:
In fact we had to revert to a version before your PR, because we found out that the generated js client for the ocis settings service (placed in `vendor/settingsClient.js`) needed the global interceptor. Still trying to figure out how to change that, because originally we didn't want to modify that generated client.

pascalwengerter:
> Fully agreed, deserves tagging. At the moment releases in this repository happen automatically by merging to master. The release on npmjs.com will have the version `1.0.0-<some-automatically-increasing-number>`. Turned out just recently (last week) that this is horrible for reverting to another version of the sdk. For now you can only have a look at the `Versions` page on npmjs.com and guesstimate if the version 2296 (which says `2 months ago`) does not include your PR, yet.

[...]

PR #855, opened 2021-05-28
Prevent owncloud-sdk from modifying global axios interceptors and cleanup upon logout

rpocklin:
I have raised issue https://github.com/owncloud/owncloud-sdk/issues/854 which this PR is a solution for.

This PR moves the interceptor modification to an sandboxed instance of **axios** so it won't affect anything not using this specific instance, which then allows for multiple instances of the `new Owncloud()` to be instantiated without affecting each other.

Additionally this ensures once the user has logged out the interceptor is unbound so that there are no memory leaks and the instance of `this.axios` can be garbage collected properly.


kulmann (approved):
Nice! Thanks for taking care. LGTM 🚀

PR #859, opened 2021-05-31
Bump pact-js to 10.0.0-beta.36

dpakach:
Bump pact-js to 10.0.0-beta.36

- Use matchers  in XmlBuilder instead of string literals


individual-it (approved):
that issue should be fixed with the next release.

PR #868, opened 2021-06-22
Revert "Prevent owncloud-sdk from modifying global axios interceptors and cleanup upon logout"

pascalwengerter:
Reverts owncloud/owncloud-sdk#855


fschade (approved):
🚀 LGTM to revert for now, but we should check whats wrong here, me like interceptors muchos ;)

PR #872, opened 2021-06-23
Revert "Revert "Prevent owncloud-sdk from modifying global axios interceptors and cleanup upon logout""

pascalwengerter:
Reverts owncloud/owncloud-sdk#868 / Re-introduces #855 because it's actually a neat improvement


kulmann (approved):
Thanks for taking care 👍

PR #919, opened 2021-10-19
Allow cache control when getting a file

diocas:
If I edit a file with the markdown editor (or our CERN fork), the PUT request it does receives a reply (at least from OCIS) that instructs the browser to cache the output. This means that if I open the file again, I might not be receiving the latest content that is actually stored in the FS.

I had a quick discussion with Benedikt and proposed allowing a no-cache request.
Here is a possible implementation but please let me know if you have any other idea.

### DIFF
--- a/src/fileManagement.js
+++ b/src/fileManagement.js
@@ -69,7 +69,7 @@ class Files {
    * @returns {Promise.<error>}       string: error message, if any.
    */
   getFileContents (path, options = {}) {
-    return this.helpers._get(this.helpers._buildFullWebDAVURL(path)).then(data => {
+    return this.helpers._get(this.helpers._buildFullWebDAVURL(path), options).then(data => {
       const response = data.response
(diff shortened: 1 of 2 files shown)


diocas:
@kulmann ok. Any prediction for when this will be fixed?
Maybe at CERN we can use a newer version of the SDK and disable the settings.

kulmann:
We'll need it soonish, but will take some care of this repo before we update in `web`. For now probably a better strategy if you use this repo master in your fork.

Is this PR ready for merge or do you still want to change things?

pascalwengerter:
> Let's see of #925 fixes it already. @kulmann could you review? Upon merge we'll see the result:)

Didn't work, see [here](https://drone.owncloud.com/owncloud/owncloud-sdk/2733/5/5). Will gather more insights what could have gone wrong, maybe outdated GH api keys?

PR #926, opened 2021-10-20
Bump oCIS commit id

pascalwengerter:
Supersedes #900


individual-it:
one problem found is that current ocis returns 400 when requested PROPFIND with `Depth: 2`. This is NOT a bug in ocic because `Depth: 2` is not valid, still the older version of ocis and also cC10 are not so fussy, so the issue was not noticed.
For more details see: https://github.com/owncloud/ocis/issues/1975

individual-it:
tests don't crash anymore, now need to find out what changed in ocis and adjust the tests

kulmann (approved):
finally 🤩

PR #933, opened 2021-11-03
Cache getCapabilities

diocas:
Calling the capabilities endpoint is expensive. When running in CERNBox it's extremely expensive...
Since in Web we call `login()` and `getCapabilities()` one after the other, I don't see a reason to do 2 requests.

In this PR I create a cache that stores the latest retrieved value. From the web I will set `refresh` as `false`.
I implemented this way to keep the same behavior by default, but maybe the login could also take it in consideration (in case in the future we switch the order and call `getCapabilities()` first.

Let me know if you agree with this or if you prefer something else.

### DIFF
--- a/src/owncloud.js
+++ b/src/owncloud.js
@@ -81,6 +81,7 @@ class ownCloud {
         return helpers.ocs(options)
       }
     }
+    this.cache = {}
   }
 
   /**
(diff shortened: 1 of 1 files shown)


kulmann:
@diocas good improvement 👍 CI is failing, will you look into it?

kulmann (approved):

PR #937, opened 2021-11-23
Use custom url to avoid encoded by the webdav library

diocas:
We implemented the recycle bin of our EOS projects with a query parameter. This was discussed with the OCIS backend team and they will use the same way to return the recycle of spaces.

The `webdav` client encodes the path, unless if we manually set `url` in the request options. In this PR I generate the URL in the SDK instead of relying in the library.

What do you think?
Btw, to set the recycle query param we just set as the url... We could provide an extra option in this function call if you prefer.

### DIFF
--- a/src/dav.js
+++ b/src/dav.js
@@ -15,6 +15,7 @@ export class Dav {
     }
 
     this.baseUrl = baseUrl
+    this.baseUrlv2 = baseUrlv2
 
     this.userName = null
 
@@ -216,16 +217,22 @@ export class Dav {
   request (method, path, headers, body, options = {}) {
     options.version = options.version || 'v1'
     const reqClient = options.version === 'v2' ? this.clientv2 : this.client
+    const reqBaseUrl = options.version === 'v2' ? this.baseUrlv2 : this.baseUrl
     delete options.version
 
+    const params = new URLSearchParams(options.query).toString()
+    const query = params ? '?' + params : ''
+    delete options.query
+
     const requestOptions = {
+      url: reqBaseUrl + path + query,
       method,
       headers,
       data: body,
       ...options
     }
     return new Promise((resolve) => {
-      return reqClient.customRequest(decodeURIComponent(path), requestOptions)
+      return reqClient.customRequest('', requestOptions)
         .then(res => {
           var resultBody = res.data
           if (res.status === 207) {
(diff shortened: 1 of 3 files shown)


diocas:
@kulmann can you, or someone else, have a look?

kulmann:
The test suite doesn't seem to like it. Could someone from Jankaritech take a look please @phil-davis @individual-it ? Thanks!

diocas:
@kulmann @phil-davis @individual-it any update? If not possible to merge, I will have to fork for now.

kulmann:
> @kulmann @phil-davis @individual-it any update? If not possible to merge, I will have to fork for now.

pinged in internal chat again

phil-davis:
> pinged in internal chat again

@diocas @kulmann sorry, we missed responding to this. I have put this in Fastlane of our QA project. One of the guys who usually works with this test suite will have a look first thing in the morning our time (10pm here now). I don't have enough current experience with this to be useful at 10pm!

src/dav.js
```diff
@@ -15,6 +15,7 @@ export class Dav {
     }
 
     this.baseUrl = baseUrl
+    this.baseUrlv2 = baseUrlv2
```
individual-it, line 18: should we then call `baseUrl` `baseUrlv1` ?

individual-it (changes requested):

dpakach:
I'll have a look at this

src/dav.js
diocas: I would say it's consistent this way, since you also have `client` and `clientv2`. I can change it, but it would need to change in many places.

kulmann: also in favour of keeping `baseUrl` for the first one.

kulmann:
Thank you @dpakach and @phil-davis !

dpakach:
@kulmann @diocas I quite don't understand how this is supposed to work. I tried to implement this but ran into a problem.
for eg. `new?folder` is a valid file/folder name in owncloud. now if you would like to request to this endpoint with an additional parameter then you will have to request with path value like `/new?folder?query=value` and this cannot be parsed.
Now either the client who calls the function `oc.files.list()` should themself encode the path and call the function like 
```
oc.files.list('new%3Ffolder?query=value')
```
I don't think this would work because this would break the current version(and also is not a good solution to make the client encode it).
or we should add an additional parameter to pass the query parameters for the clients.

diocas:
Good point... Let me try to fix it. We can sill encode everything and use query parameters as a separate and optional parameter of the function.

kulmann (approved):
Thanks for taking care!

individual-it (approved):

diocas:
We still don't have a test for the query parameter, but I guess we can still merge it?

individual-it:
merged, here an issue to add tests #942

Issue #942, opened 2021-12-07
create tests for query parameter for trashbin


individual-it:
issue: #936 
fix: #937

PR #943, opened 2021-12-09
Query parameters for all recycle functions

diocas:
I forgot that besides listing the recycle, users might also want to "restore"... Who knew...


diocas:
@kulmann @individual-it

individual-it (approved):

individual-it:
tests to be added in https://github.com/owncloud/owncloud-sdk/issues/942

Issue #958, opened 2022-01-24
Drop support for webdav v1


kulmann:
We're currently planning to drop support for webdav (v1) in favour of the new dav implementation (v2), which is fully supported by both ownCloud 10 and ownCloud Infinite Scale.

Would like to hear if there are objections to this.

kulmann:
cc @rpocklin

kulmann:
cc @phil-davis @individual-it

phil-davis:
I don't see a problem with that. All server implementations have the "new" webdav v2. So clients and client infrastructure does not have to know anything about "old" webdav v1.

fschade:
Is there a reason why ownCloud classic to still supports webdav (v1)? If i get it right dav (v2) has the same set of features!?

So if we do not miss something i vote to drop it.

DeepDiver1975:
v1 only serves the files of the currently logged in user
v1 serves files, calendar, contacts, trashbin and many more following the URI principle that URIs are truly unique ...

remote.php/webdav/welcome.txt is not unique as it depends on the logged in user
remote.php/dav/files/alice/welcome.txt in contrary is unique 

the over all plan was once to even allow cross-user operations like copying files from one user space to another .... but we never got there ...

DeepDiver1975:
> So if we do not miss something i vote to drop it.

from the sdk perspective there is no reason to stay with v1 - it should have been used right from the beginning :see_no_evil:

JanAckermann:
@DeepDiver1975 Thx, so we will proceed with replace v1 with v2

JanAckermann:
@butonic But if we release a new major version, this might ain’t an issue ?

pascalwengerter:
> @butonic But if we release a new major version, this might ain’t an issue ?

We should switch the whole thing over to the way we do semver in the other web-related repos I suggest :)

kulmann:
> Um, might be needed for cernbox @kulmann

@diocas @elizavetaRa do you maybe know?

diocas:
For CERNBox should not matter, assuming that OCIS has the correct implementation.

PR #961, opened 2022-01-25
Update pact lib to v10.0.0-beta.54

JammingBen:
fixes https://github.com/owncloud/owncloud-sdk/issues/959

@pascalwengerter fyi

PR #962, opened 2022-01-25
Change to webDavV2 + cleanup

JanAckermann:
fixes https://github.com/owncloud/owncloud-sdk/issues/958

### DIFF
--- a/src/dav.js
+++ b/src/dav.js
@@ -5,7 +5,7 @@ const applicationXmlResponseHeaders = {
   'Content-Type': 'application/xml; charset=utf-8'
 }
 export class Dav {
-  constructor (baseUrl, baseUrlv2) {
+  constructor (baseUrl) {
     this._XML_CHAR_MAP = {
       '<': '&lt;',
       '>': '&gt;',
@@ -15,7 +15,6 @@ export class Dav {
     }
 
     this.baseUrl = baseUrl
-    this.baseUrlv2 = baseUrlv2
 
     this.userName = null
 
@@ -32,7 +31,6 @@ export class Dav {
     }
 
     this.client = createClient(baseUrl, {})
-    this.clientv2 = createClient(baseUrlv2, {})
   }
 
   _escapeXml (s) {
@@ -206,7 +204,7 @@ export class Dav {
    * Performs a HTTP request, and returns a Promise
    *
    * @param {string} method HTTP method
-   * @param {string} url Relative or absolute url
+   * @param {string} path Relative or absolute url
    * @param {Object} headers HTTP headers as an object.
    * @param {string} body HTTP request body.
    * @param {Object} options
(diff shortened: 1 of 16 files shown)


pascalwengerter (commented):
Please change the version number in package.json to `2.0.0-alpha` (or similiar) if it's a breaking change and add a changelog item (after rebase on master which now is on the usual semver/calens-changelog flow)

src/dav.js
```diff
@@ -211,14 +211,11 @@ export class Dav {
...
    * @param {Function} options.onProgress progress callback
-   * @param {davVersion} options.version v1 | v2
    * @return {Promise}
    */
   request (method, path, headers, body, options = {}) {
-    options.version = options.version || 'v1'
-    const reqClient = options.version === 'v2' ? this.clientv2 : this.client
-    const reqBaseUrl = options.version === 'v2' ? this.baseUrlv2 : this.baseUrl
-    delete options.version
+    const reqClient = this.clientv2
```
kulmann, line 217: All the logic with `baseUrl` and `client` (v1) can be dropped from the codebase, right? It's a 2.0.0 release anyway so I'd vote for cleaning it up.

src/fileManagement.js
```diff
@@ -91,24 +102,15 @@ class Files {
...
-  }
-
   /**
    * Returns the url of a remote file - using the version 2 endpoint
    * @param   {string}  path    path of the remote file at OC instance
    * @returns {string}          Url of the remote file
    */
-  getFileUrlV2 (path) {
+  getFileUrl (path) {
+    path = this._sanitizePath(path)
```
kulmann, line 111:
```suggestion
    path = '/' + this._sanitizePath(path)
```
and drop the next line, since you implemented `_sanitizePath` to remove leading slashes.

src/dav.js
JanAckermann: will do thx

fschade:
Please do not merge before tomorrow, i wonder why the tests still pass even if we nearly changed all routes.

cc.: @individual-it, can you give us a short onboarding for pact tomorrow?

JammingBen:
> i wonder why the tests still pass even if we nearly changed all routes.

Because we also adjusted the tests 😄  Basically all `/remote.php/webdav/...` -> `/remote.php/dav/files/${testUser}/...`

fschade:
nice, thanks for bringing light into the dark :) so lets see what next test run says

src/dav.js
fschade: done

src/fileManagement.js
fschade: done

fschade:
ci fails now, alpha release postponed to tomorrow

JammingBen:
It fails because you've updated pact to `10.0.0-beta.54`. We cannot use that version (yet at least), see https://github.com/owncloud/owncloud-sdk/pull/961#issuecomment-1021064665.

fschade (approved):
🚀

JanAckermann:
> (after rebase on master which now is on the usual semver/calens-changelog flow)

Aaaaand it was merged without the changelog :D

PR #964, opened 2022-01-26
Introduce semver

pascalwengerter:
Switching the SDK to [semver](https://semver.org/), adding the changelog mechanism we use in most/all other ownCloud repos, adding SonarCloud and removing the pre-commit hook


kulmann (approved):
Awesome! LGTM 👍 🚀

PR #977, opened 2022-02-01
Reject promise when network request fails in OCS

kulmann:
branch is based on tag v1.1.1 so that we can make a v1.1.2 from it. but can be merged back to master after release to get the fix into that as well. Needs to drop the version bump again of course. Or just cherry pick the first commit to master...


kulmann:
Needed to unblock https://github.com/owncloud/web/pull/6309

kulmann:
Closing here because the CI for release 1.1.2 succeeded, see https://github.com/owncloud/owncloud-sdk/releases/tag/v1.1.2 and the commit has been brought back to master via https://github.com/owncloud/owncloud-sdk/pull/978

PR #978, opened 2022-02-01
Reject promise when network request fails in OCS

kulmann:
Same as #977 but without the version bump to v1.1.2 (because master already is at v2.0.0-alpha.x).


pascalwengerter (approved):
LGTM 🐛

PR #979, opened 2022-02-03
Change: Move dependencies to peerDependencies

### LINKED ISSUE #972: Make axios a peer-dependency
Currently in https://github.com/owncloud/web, the main user of the SDK, we're also using axios for other stuff and end up with two axios-es in our bundle. It might be smart to make axios a peer-dependency in the SDK so `web` only needs to ship with one of them :)

pascalwengerter:
Suggestion to align the used versions of named packages (for the `filepicker` and `web`) and reduce the bundle size (e.g. currently, `web` ships with two different axios versions in its bundle)

Fixes #972 

~Please don't merge, we'll have to test for capability of version ranges first~ should be good to go since https://github.com/owncloud/web/pull/6287 has a green web CI with the previous breaking changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,3 @@
-[![Build Status](https://travis-ci.org/owncloud/owncloud-sdk.svg?branch=master)](https://travis-ci.org/owncloud/owncloud-sdk)
-[![codecov](https://codecov.io/gh/owncloud/owncloud-sdk/branch/master/graph/badge.svg)](https://codecov.io/gh/owncloud/owncloud-sdk)
 [![docs](https://img.shields.io/badge/api_docs-online-blue.svg)](https://owncloud.github.io/owncloud-sdk/)
 
 # ownCloud JavaScript SDK
--- a/README.md
+++ b/README.md
@@ -20,6 +20,18 @@ yarn add owncloud-sdk
 
 to add the `owncloud-sdk` to your project.
 
+If you haven't done so already, you also need to add the following `peerDependencies` to your dependencies:
+
+```
+npm install axios cross-fetch promise qs semver utf8 uuid webdav xml-js
+```
+
+or
+
+```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,5 +1,4 @@
 [![Build Status](https://travis-ci.org/owncloud/owncloud-sdk.svg?branch=master)](https://travis-ci.org/owncloud/owncloud-sdk)
```
kulmann, line 1: travis CI ref should also be removed, since the last build ran 3 years ago.

changelog/unreleased/change-peerdependencies
```diff
@@ -0,0 +1,9 @@
+Change: Use peerDependencies instead of dependencies
+
+In the past, we used dependencies in package.json which lead to a big bundle size and increased the possibility 
+that the same package with 2 different versions is part of the final bundle of the consuming application.
+
+From now on, dependencies that are required to use the SDK are added to the peerDependencies section in package.json.
```
kulmann, line 6: It's common practice to mention the peer dependencies in the install commands in the `Readme.md` as well. Since we list install commands there I'd appreciate to see them added there for us as well.

kulmann (changes requested):

README.md
pascalwengerter: Aye 👀

changelog/unreleased/change-peerdependencies
pascalwengerter: We should do the same in the ODS, then? But good point, working on it

kulmann: yep, that would be a nice improvement 👍

kulmann: e.g. https://www.npmjs.com/package/vue-datetime#bundler-webpack-rollup

kulmann (approved):
Nice 💪

PR #985, opened 2022-02-07
Fix file path stripping for favorites, search and tags

JammingBen:


JanAckermann (approved):

pascalwengerter (approved):
Curious why CI didn't complain when you did the initial changes, you got any insights on that? :)

JammingBen:
I think because the SDK stripped away `/files/username` form the response body, and the tests expected exactly that.

PR #1002, opened 2022-02-14
Restore query parameters on recycle list

diocas:
Remove no longer needed dav version option

Looks like if you removed one extra line (and one less) when removing the old dav endpoints.

(ping @pascalwengerter @kulmann)

### DIFF
--- a/src/filesTrash.js
+++ b/src/filesTrash.js
@@ -45,7 +45,10 @@ class FilesTrash {
       this.helpers._buildFullDAVPath(target),
       properties,
       depth,
-      headers
+      headers,
+      {
+        query: query
+      }
     ).then(result => {
       if (result.status !== 207) {
         return Promise.reject(this.helpers.buildHttpErrorFromDavResponse(result.status, result.body))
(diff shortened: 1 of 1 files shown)


src/filesTrash.js
```diff
@@ -45,7 +45,10 @@ class FilesTrash {
       this.helpers._buildFullDAVPath(target),
       properties,
       depth,
-      headers
+      headers,
+      {
+        query: query
```
kulmann, line 50:
```suggestion
        query
```

kulmann (changes requested):
Hmm, sorry that it got lost. 😞 One nitpick, otherwise of course good to go

src/filesTrash.js
diocas: Done

diocas:
@kulmann can we merge?

kulmann (approved):
👍

diocas:
Are you still building the dev versions?

kulmann:
> Are you still building the dev versions?

No, there is semantic versioning with manual releases now in this repo. I will create one for you.

diocas:
thanks

PR #1013, opened 2022-02-18
Adjust share management to properly work with spaces

JammingBen:


changelog/unreleased/change-share-spaces
```diff
@@ -0,0 +1,5 @@
+Change: Adjust share management to properly work with spaces
```
kulmann, line 1: Why is this not an enhancement?

JammingBen: It is 😄  Thx for spotting!

src/shareManagement.js
```diff
@@ -123,6 +124,49 @@ class Shares {
...
+   */
+  shareSpaceWithUser (path, username, spaceId, optionalParams) {
+    const postData = {
+      shareType: this.helpers.OCS_SHARE_TYPE_SPACE,
+      shareWith: username,
+      space_ref: spaceId
+    }
+
+    if (path) {
+      postData.path = this.helpers._normalizePath(path)
```
kulmann, line 144: Do I get it right that a space share is either the entire space or a specific file from the space?

kulmann (changes requested):
`shareFileWithUser` and `shareSpaceWithUser` are very similar. Can you extract generic parts?

src/shareManagement.js
JammingBen: Honestly, I haven't dealt with sharing resources within a space yet. I assumed it works the same, so I left the path. I'll gather some information about this

JammingBen: As discussed: we keep it as it is for now

kulmann (approved):

PR #1016, opened 2022-03-03
Add X-Request-ID to headers unconditionally

kulmann:
Add X-Request-ID to headers unconditionally. The version we were requiring already is ages old, doesn't seem relevant anymore. And obviously preventing us to let oCIS announce its correct version, see https://github.com/owncloud/ocis/pull/2918

Removing the version check also meant not needing the semver package
anymore. Removed it from devDependencies, peerDependencies and
documentation accordingly.


fschade (approved):
LGTM

PR #1020, opened 2022-03-04
Add X-Requested-With header to all requests

dschmidt:
oC 10 uses this header to determine whether a request is an ajax
request. If we want to get rid of this, we need to make sure it's
not used anymore. For example currently it's used to determine
whether a PROPFIND on a password protected link should show
an auth popup or not.

Needed to fix https://github.com/owncloud/web/issues/5727


kulmann (approved):
👏

PR #1021, opened 2022-03-08
Require path for FilesTrash bin class

JanAckermann:
Change: Pass full trash bin path to methods of FilesTrash class

Since incompatibility with spaces, we changed the way how the methods of the class
FilesTrash(filesTrash.js) need to be called.
Now it is mandatory to pass the full webDav(v2) path of the trash bin.

### DIFF
--- a/src/filesTrash.js
+++ b/src/filesTrash.js
@@ -105,7 +105,7 @@ class FilesTrash {
    */
   restore (trashBinPath, fileId, destinationPath, overWrite = false, query = undefined) {
     if (trashBinPath === undefined) {
-      return Promise.reject(new Error('No tashBinPath given for restore'))
+      return Promise.reject(new Error('No trashBinPath given for restore'))
     }
     if (fileId === undefined) {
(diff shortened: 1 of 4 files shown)


src/filesTrash.js
```diff
@@ -94,25 +96,31 @@ class FilesTrash {
...
-   * @param   {string}        originalLocation
+   * @param   {string}        destinationPath
    * @param   {boolean}       overWrite
    * @param   {Object}        query            Query parameters dictionary
    * @returns {Promise<void | string | Error>}
    */
-  restore (fileId, originalLocation, overWrite = false, query = undefined) {
+  restore (trashBinPath, fileId, destinationPath, overWrite = false, query = undefined) {
+    if (trashBinPath === undefined) {
+      return Promise.reject(new Error('No tashBinPath given for restore'))
```
kulmann, line 108:
```suggestion
      return Promise.reject(new Error('No trashBinPath given for restore'))
```

JammingBen (approved):
LGTM!

PR #1025, opened 2022-03-11
Add the 'space_ref' param to the share management

JammingBen:
Implement the `space_ref` param according to https://github.com/cs3org/reva/pull/2622.

Needed for https://github.com/owncloud/web/pull/6577/files.

### DIFF
--- a/src/shareManagement.js
+++ b/src/shareManagement.js
@@ -90,18 +90,26 @@ class Shares {
    */
   shareFileWithUser (path, username, optionalParams) {
     path = this.helpers._normalizePath(path)
+    let urlPath = 'shares'
 
     let postData = {
       shareType: this.helpers.OCS_SHARE_TYPE_USER,
-      shareWith: username,
-      path: path
+      shareWith: username
+    }
+
+    if (!optionalParams || !optionalParams.spaceRef) {
+      postData.path = path
     }
 
     if (optionalParams) {
+      if (optionalParams.spaceRef) {
+        urlPath += `?space_ref=${optionalParams.spaceRef}`
+      }
+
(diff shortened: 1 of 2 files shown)


src/shareManagement.js
```diff
@@ -90,18 +90,26 @@ class Shares {
...
+      postData.path = path
     }
 
     if (optionalParams) {
+      if (optionalParams.spaceRef) {
+        urlPath += `?space_ref=${optionalParams.spaceRef}`
+      }
+
       postData = { ...postData, ...this._getOptionalParams(optionalParams) }
     }
```
JanAckermann, line 111: a combination of Lodash _.snakeCase() and  https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams might be more programmatic

JammingBen: I like the idea with the `URLSearchParams` and implemented those 👍 
However, I prefer keeping `space_ref` hardcoded because `spaceRef` is basically just a variable naming, so I don't like those to be dependent on each other.

src/shareManagement.js
```diff
@@ -90,18 +90,28 @@ class Shares {
...
+    const urlParams = new URLSearchParams()
 
     let postData = {
       shareType: this.helpers.OCS_SHARE_TYPE_USER,
-      shareWith: username,
-      path: path
+      shareWith: username
+    }
+
+    if (!optionalParams || !optionalParams.spaceRef) {
```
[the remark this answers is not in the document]
JammingBen: I originally had that 😂 Decided against it because I found deleting it is not so nice as well. But no hard feelings here, I can revert it tomorrow.

kulmann (approved):

PR #1032, opened 2022-04-01
Bugfix propfind response cut off

lookacat:
We've fixed an issue where `_parsePropNode` in the dav client cut off subobjects of an object.

### DIFF
--- a/src/dav.js
+++ b/src/dav.js
@@ -282,6 +282,19 @@ export class Dav {
           }
           continue
         }
+        if (node.constructor === Object && Object.keys(node).length !== 0) {
+          const parseChildren = (parseNode) => {
+            Object.entries(parseNode).forEach(([key, value]) => {
+              delete parseNode[key]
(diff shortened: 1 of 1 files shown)


pascalwengerter:
Tried running the failing oCIS Providertests locally

Steps:
- Start oCIS from _around the commitid linked in .drone.env_ in the web repo (updated docker-compose and run `docker compose up ocis`)
- Run `PROVIDER_BASE_URL=https://host.docker.internal:9200 yarn test-provider:ocis` which fails on Apple M1 chips since the @pact-foundation lib doesn't support it yet. I'm sure there's more env variables missing (further investigation in the .drone.star needed). The way to go here would be having them run successfully locally (= green tests) on master and then switch to this branch/PR and fix them

lookacat:
We won't continue here since it's too time consuming and we are not happy with the parsing overall.
Will discussed with:
https://github.com/owncloud/owncloud-sdk/issues/1036

PR #1035, opened 2022-04-04
Replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/changelog/unreleased/change-replace-deprecated-substr
+++ b/changelog/unreleased/change-replace-deprecated-substr
@@ -1,0 +1,7 @@
+Change: Replace deprecated String.prototype.substr()
+
+We've replaced all occurrences of the deprecated String.prototype.substr()
+function with String.prototype.slice() which works similarly but isn't
+deprecated.
+
+https://github.com/owncloud/owncloud-sdk/pull/1035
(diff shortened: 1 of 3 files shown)


changelog/unreleased/change-replace-deprecated-substr
```diff
@@ -0,0 +1,7 @@
+Change: Replace deprecated String.prototype.substr()
```
pascalwengerter, line 1:
```suggestion
Enhancement: Replace deprecated String.prototype.substr()
```

Since this isn't a user-facing breaking change it counts as `enhancement` in SemVer, could you update the changelog item filename & heading?:)

pascalwengerter (commented):
Thanks for opening this PR! Only a minor nitpick regarding the changelog

changelog/unreleased/change-replace-deprecated-substr
CommanderRoot: Sure. I updated the file as requested

pascalwengerter (approved):
LGTM, thanks for taking care

PR #1038, opened 2022-04-06
update go since 1.16 is eol

wkloucek:


pascalwengerter (approved):

pascalwengerter:
Closing as included in #926 which also bumps the commit ID (which is necessary for the higher Go version with it's depndencies, I reckon?)

PR #1042, opened 2022-04-14
Respect optional quicklink param for shareFileWithLink

fschade:
[#2715](https://github.com/cs3org/reva/pull/2715) introduced a new link variant which only can exist once per resource, this is needed for [6605](https://github.com/owncloud/web/issues/6605).

### DIFF
--- a/changelog/unreleased/enhancement-quicklink
+++ b/changelog/unreleased/enhancement-quicklink
@@ -1,0 +1,8 @@
+Enhancement: Create quicklink
+
+We've added the option to create quicklinks which under the hood is the same as a link expect
+the fact that a quicklink can only exist once per item.
+
+https://github.com/owncloud/owncloud-sdk/pull/1041
+https://github.com/owncloud/web/issues/6605
+https://github.com/cs3org/reva/pull/2715
--- a/changelog/unreleased/enhancement-quicklink
+++ b/changelog/unreleased/enhancement-quicklink
@@ -1,8 +1,8 @@
 Enhancement: Create quicklink
 
-We've added the option to create quicklinks which under the hood is the same as a link expect
-the fact that a quicklink can only exist once per item.
+We've added the option to define if a share link is of type quicklink or not.
 
 https://github.com/owncloud/owncloud-sdk/pull/1041
 https://github.com/owncloud/web/issues/6605
+https://github.com/owncloud/core/pull/39167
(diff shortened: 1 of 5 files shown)


changelog/unreleased/enhancement-quicklink
```diff
@@ -0,0 +1,8 @@
+Enhancement: Create quicklink
+
+We've added the option to create quicklinks which under the hood is the same as a link expect
```
kulmann, line 3:
```suggestion
We've added the option to create quicklinks which under the hood is the same as a link except for
```

src/shareManagement.js
```diff
@@ -74,6 +73,9 @@ class Shares {
       if (optionalParams.attributes) {
         postData.attributes = optionalParams.attributes
       }
+      if (optionalParams.quicklink) {
+        postData.quicklink = optionalParams.quicklink
```
kulmann, line 77: isn't the quicklink flag supposed to go into the `attributes` (see above)? Otherwise we couldn't have it for oc10, because an additional `quicklink` prop wouldn't be handled by the backend.

fschade: we've decided to keep it on the top level instead of putting it into the attributes (cs3).
For oc10 we decided to catch the quicklink param here and mirrow it into the attributes.

kulmann: how do you decide which one to do (cs3 / oc10)? also, for a migration from oc10 to ocis this means that we need to keep this in mind particularly, instead of just blindly copying all attributes of a link.

fschade: i do not differentiate the backends, just send both... and add a comment why thats the case... we talked about it this morning and we all agreed `attributes` is more or less like a param blackbox. Having it top-level is easier to document and even more clean.

Are you against it?

kulmann: sending both is fine then. wanted to make sure that we don't have oc10/ocis if/else constructs in the sdk code.

this still has the drawback of having to write the toplevel quicklink flag during migration from oc10 to ocis. please make sure to add that to some migration considerations collection and then go ahead like you decided. 👍

fschade: will do, thanks for your input. Ill jump back into this pr right after we have merged the web groundwork pr.

kulmann (approved):

PR #1046, opened 2022-04-25
Manipulate share type

JanAckermann:


src/shareInfo.js
```diff
@@ -5,6 +5,14 @@
  * @classdesc ShareInfo class, stores information regarding a share
  * @param {object} ShareInfo containing information like id, url etc. of the share
  */
+
+const USER_TYPE_USER = 0
+const USER_TYPE_GUEST = 1
+const SHARE_TYPE_USER = 0;
+const SHARE_TYPE_GROUP = 1;
+const SHARE_TYPE_LINK = 3;
+const SHARE_TYPE_GUEST = 4;
```
kulmann, line 14: I think it's sufficient to only introduce the constant that you need and link to the documentation for other share types (same link for user type and ping the docs team to add the user type prop to the docs): https://doc.owncloud.com/server/next/developer_manual/core/apis/ocs-share-api.html#response-attributes-2

kulmann (commented):
Code itself looks good. But the linter is screaming and the changelog is empty. 😁

JanAckermann:
> Code itself looks good. But the linter is screaming and the changelog is empty. grin

DRAFT DRAFT DRAFT :D :D :D 
I even wasn't able to check if it works, as stuff in the other pr is failing :P

src/shareInfo.js
JanAckermann: https://github.com/owncloud/docs-server/issues/272#issuecomment-1108501320

src/shareInfo.js
```diff
@@ -5,17 +5,21 @@
  * @classdesc ShareInfo class, stores information regarding a share
  * @param {object} ShareInfo containing information like id, url etc. of the share
  */
-class ShareInfo {
-  constructor (shareInfo) {
-    this.shareInfo = {}
 
-    // Below keys don't need to be stored
-    const notNeededKeys = ['item_type', 'item_source', 'file_source', 'parent', 'storage', 'mail_send']
```
JanAckermann, line 13: we need the storage, and I have no idea, why we should purge those properties

changelog/unreleased/enhancement-enforce-share-type-guest-if-applies
```diff
@@ -0,0 +1,7 @@
+Enhancement: Enforce `share_type` guest if applies
+
+With oC10, the `share_type` is not set to guest, if a user shares a resource with a guest user,
+therefore we check check if the property `share_with_user_type` is present and set to `guest`,
```
kulmann, line 4:
```suggestion
therefore we check if the property `share_with_user_type` is present and set to `guest`,
```

src/shareInfo.js
```diff
@@ -5,17 +5,21 @@
...
-    for (const key in shareInfo) {
-      if (!(key in notNeededKeys)) {
-        this.shareInfo[key] = shareInfo[key]
-      }
+class ShareInfo {
+  constructor (shareInfo) {
+    this.shareInfo = shareInfo
+
+    /**
+     * oC10 does not emit if a share receiver is a user or a guest,
```
kulmann, line 17: I'd like to have this comment less technical, e.g. like this:
```
oc10 does not use the share type `guest` but `user` for guest shares. As mitigation it emits `share_with_user_type` with value `0` for user and `1` for guest shares. We utilize this information to rewrite the share_type to guest when needed.
```

src/shareManagement.js
```diff
@@ -458,7 +469,20 @@ class Shares {
         throw new Error(response.status + '/' + response.statusText)
       })
       .then(json => {
-        return json.ocs.data
+        const data = json.ocs.data
+
+        /**
+         * oC10 does not emit if a share recipient is a user or a guest,
```
kulmann, line 475: same as above for simplification of the comment.

kulmann (approved):
nitpicks regarding changelog and code comments. the code looks good 👍

PR #1058, opened 2022-05-02
use also commit id of ocis when running tests

individual-it:
the commitid was ignored and the top of master was always used


pascalwengerter (approved):

phil-davis:
Also note: core PR https://github.com/owncloud/core/pull/40016 changed the default for `dav.propfind.depth_infinity`
So that is why we explicitly enable it here - to be sure that any depth infinity PROPFINDs that the tests may do will work in the oC10 pipeline.

PR #1059, opened 2022-05-02
Use oCIS commit ID in CI

pascalwengerter:


pascalwengerter:
Closed as duplicate of #1058

PR #1061, opened 2022-05-03
[tests-only] use latests as OC10 version

individual-it:
using the nightly build might be dangerous and is not needed, better to test against the version we would really use the SDK with


phil-davis (approved):
Looks good if CI passes.

We can leave the depth-infinity config command in the code, because it exists and works in the current latest oC10, which is 10.9.1, and it will be really needed when 10.10.0 is released.

pascalwengerter:
Thanks for taking care here!

PR #1062, opened 2022-05-03
[tests-only] delete test for PROPFIND depth 2

individual-it:
depth 2 with PROPFIND is not a valid thing in WebDAV, so delete the test


phil-davis (approved):
For information: http://www.webdav.org/specs/rfc4918.html#HEADER_Depth
```
Depth = "Depth" ":" ("0" | "1" | "infinity")
```

PR #1073, opened 2022-05-25
Add overwrite flag for move

lookacat:
### Description:

We've added an overwrite flag for the file move operation

### DIFF
--- a/src/fileManagement.js
+++ b/src/fileManagement.js
@@ -253,7 +253,7 @@ class Files {
    * @returns {Promise.<status>}  boolean: whether the operation was successful
    * @returns {Promise.<error>}   string: error message, if any.
    */
-  move (source, target, overWrite=false) {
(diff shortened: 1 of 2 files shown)


src/fileManagement.js
```diff
@@ -253,7 +253,7 @@ class Files {
    * @returns {Promise.<status>}  boolean: whether the operation was successful
    * @returns {Promise.<error>}   string: error message, if any.
    */
-  move (source, target) {
+  move (source, target, overWrite = false) {
```
kulmann, line 256: please don't use camelcase - `overwrite` is one word, all lower case :-)

lookacat: ah sorry was copy pasta :D thx

lookacat: done

kulmann (approved):

PR #1079, opened 2022-05-30
Optional no cache for public links

diocas:
Just like we added the `noCache` option to `getFileContents`, I just realized the same needs to be applied to the public links after all.

Let me know if you prefer this as a parameter or something more generic.


diocas:
@kulmann done

PR #1082, opened 2022-06-01
[tests-only] Remove passing tests from pending state (1)

saw-jan:

### Related issue

### DIFF
--- a/tests/capabilitiesTest.js
+++ b/tests/capabilitiesTest.js
@@ -14,7 +14,7 @@ describe('Main: Currently testing getConfig, getVersion and getCapabilities', ()
   } = require('./helpers/pactHelper.js')
 
   it('checking method : getConfig', async () => {
-    const provider = createProvider(false, false)
+    const provider = createProvider()
     await getCapabilitiesInteraction(provider)
     await getCurrentUserInformationInteraction(provider)
     await provider
@@ -52,7 +52,7 @@ describe('Main: Currently testing getConfig, getVersion and getCapabilities', ()
   })
 
   it('checking method : getCapabilities', async () => {
-    const provider = createProvider(false, false)
+    const provider = createProvider()
     await getCapabilitiesInteraction(provider)
     await getCurrentUserInformationInteraction(provider)
     await provider.executeTest(async () => {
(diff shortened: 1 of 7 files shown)


saw-jan:
Blocked until https://github.com/owncloud/owncloud-sdk/pull/1085

tests/helpers/pactHelper.js
```diff
@@ -70,7 +70,7 @@ const applicationXmlResponseHeaders = {
   'Content-Type': 'application/xml; charset=utf-8'
 }
 const textPlainResponseHeaders = {
-  'Content-Type': 'text/plain; charset=utf-8'
+  'Content-Type': MatchersV3.regex('text/plain(;()?charset=(utf|UTF)-8)?', 'text/plain; charset=utf-8')
```
saw-jan, line 73: Can we do loose match here or not acceptable?

Matches:
`text/plain; charset=utf-8`
`text/plain;charset=utf-8`
`text/plain; charset=UTF-8`
`text/plain`

tests/helpers/pactHelper.js
```diff
@@ -208,7 +208,7 @@ const deleteResourceInteraction = (
   if (resource.includes('nonExistent')) {
     response = {
       status: 404,
-      headers: applicationXmlResponseHeaders,
+      headers: xmlResponseHeaders,
```
saw-jan, line 211: Also, can we treat `application/xml` and `text/xml` as same?

individual-it: It think yes, see https://github.com/owncloud/ocis/issues/1686

tests/helpers/pactHelper.js
saw-jan: I will collect the data

tests/capabilitiesTest.js
```diff
@@ -14,7 +14,7 @@ describe('Main: Currently testing getConfig, getVersion and getCapabilities', ()
   } = require('./helpers/pactHelper.js')
 
   it('checking method : getConfig', async () => {
-    const provider = createProvider(false, true)
+    const provider = createProvider(false, false)
```
individual-it, line 17:
```suggestion
    const provider = createProvider()
```
as `false` is the default we could get rid of the parameters where-ever its false for both

individual-it (changes requested):

individual-it (approved):

Issue #1083, opened 2022-06-02
[contract-tests] Should be able to create user/group via graph-api


saw-jan:
Tests should be able to create user/group via graph-api like in `owncloud-test-middleware` (https://github.com/owncloud/owncloud-test-middleware/issues/110, 
https://github.com/owncloud/owncloud-test-middleware/pull/113)

PR #1085, opened 2022-06-03
[tests-only] Implemented graph api for latest oCIS

saw-jan:
### Description
Latest oCIS server uses graph-api so the existing tests would fail with the latest oCIS.
This PR implements helper files to use graph-api when testing against oCIS server.

### Related Issue
Closes https://github.com/owncloud/owncloud-sdk/issues/1083

Follow up issue: https://github.com/owncloud/owncloud-sdk/issues/1086


kiranparajuli589 (commented):
lgtm :+1:

individual-it (approved):

Issue #1089, opened 2022-06-08
Review and remove tests out of the pending state


saw-jan:
After bumping the oCIS commit id, maybe we can move more tests out of the pending state.

PR #1090, opened 2022-06-08
[tests-only] Remove passing tests from pending state (2)

saw-jan:

### Related issue

### DIFF
--- a/tests/helpers/pactHelper.js
+++ b/tests/helpers/pactHelper.js
@@ -55,7 +55,7 @@ const shareResponseOcsData = function (node, shareType, id, permissions, fileTar
     .appendElement('uid_file_owner', '', MatchersV3.string(config.adminUsername))
     .appendElement('displayname_file_owner', '', MatchersV3.string(config.adminUsername))
     .appendElement('path', '', MatchersV3.equal(fileTarget))
-    .appendElement('file_target', '', MatchersV3.equal(fileTarget))
+    .appendElement('file_target', '', MatchersV3.includes(fileTarget))
     .appendElement('stime', '', MatchersV3.string(Math.floor(Date.now() / 1000)))
 
   if (shareType === 3) {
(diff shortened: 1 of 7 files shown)


tests/signedUrlIntegrationTest.js
```diff
@@ -28,15 +30,13 @@ describe('Signed urls', function () {
...
-          'Content-Type': 'application/xml; charset=utf-8'
-        },
+        headers: xmlResponseHeaders,
         body: new XmlBuilder('1.0', '', 'ocs').build(ocs => {
           ocs.appendElement('meta', '', (meta) => {
             return ocsMeta(meta, 'ok', '100', 'OK')
           })
             .appendElement('data', '', (data) => {
-              data.appendElement('user', '', MatchersV3.equal(username))
+              data.appendElement('user', '', MatchersV3.string(username))
```
saw-jan, line 39: in oCIS, `user` will have userId value not a username

tests/helpers/pactHelper.js
```diff
@@ -55,7 +55,7 @@ const shareResponseOcsData = function (node, shareType, id, permissions, fileTar
     .appendElement('uid_file_owner', '', MatchersV3.string(config.adminUsername))
     .appendElement('displayname_file_owner', '', MatchersV3.string(config.adminUsername))
     .appendElement('path', '', MatchersV3.equal(fileTarget))
-    .appendElement('file_target', '', MatchersV3.equal(fileTarget))
+    .appendElement('file_target', '', MatchersV3.regex(`(\\/Shares)?\\${fileTarget}`, fileTarget))
```
individual-it, line 58: the issues with that is, if the fileTarget has special char that have some meaning in regex it does not work 
e.g.
`/test space and + and #.txt` does not match `(\/Shares)?\/test space and + and #.txt`
for that the regex should be `(\/Shares)?\/test space and \+ and #\.txt`

individual-it: so would need to adjust the regex or maybe `MatchersV3.includes` will be enough
includes | value: string | Value that must include the example value as a substring.

saw-jan: Ohh, I was wondering why all the changed passed locally but not in CI.
It looks like I have not touched the test that uses this interaction.

individual-it (approved):

PR #1091, opened 2022-06-09
[tests-only] Remove passing tests from pending state (3)

saw-jan:

### Related issue


tests/sharingTest.js
```diff
@@ -123,6 +122,8 @@ describe('Main: Currently testing file/folder sharing,', function () {
       ocs.appendElement('meta', '', (meta) => {
         ocsMeta(meta, 'ok', '100', MatchersV3.regex('(OK)?', ''))
       }).appendElement('data', '', data => {
+        // following elements are matched serially
+        // so any changes in the order (from the server) breaks the tests
```
saw-jan, line 126: some tests could pass in oCIS but due to this the tests fail.
I couldn't find the way to match them in any order.

saw-jan: the test is trying to match shares
- user share
- group share
- public share

tests/providerTest.js
```diff
@@ -185,6 +185,16 @@ describe('provider testing', () => {
         return { description: 'user created' }
       }
     },
+    'user doesn\'t exist': (setup, parameters) => {
+      if (setup) {
+        const result = deleteUser(parameters.username)
+
+        chai.assert.strictEqual(
+          result.status, 200, `deleting user '${parameters.username}' failed`
+        )
```
saw-jan, line 194:
```
Message:
      deleting user 'test123' failed: expected 204 to equal 200
```
User deletion response code can either be `200` or `204`

tests/sharingTest.js
individual-it: please open an issue and add to out ToDo list, we can work on it later

tests/providerTest.js
individual-it: I think `.include` could help
```
chai.assert.include(
          [200, 204],result.status, `deleting user '${parameters.username}' failed`
        )
```

tests/sharingTest.js
saw-jan: Raised an issue for this: https://github.com/owncloud/owncloud-sdk/issues/1094

individual-it (approved):

Issue #1086, opened 2022-06-03
Refactor tests and helpers as of follow up to PR #1085


saw-jan:
Tasks:
- [x] implement separate config for test users (#1095)
- [x] use httpHelper in webdavHelper and sharingHelper (#1096)
- [x] refactor tests to use users from user config (#1095)
- [ ] use helper functions from `providerStateHelper` for provider state management (#1098, #1099, #1100)

PR #1099, opened 2022-06-16
Replace provider states with helper functions (p2)

saw-jan:
For the Provider side tests, we have stateHandlers for pre-conditions. We have used state handlers directly in most of the tests but we also have state helper functions exactly for that.
State handler:
https://github.com/owncloud/owncloud-sdk/blob/4116766b9f90e6283d34bbb8eda2baae2ffaf02b/tests/providerTest.js#L112-L120

helper function:
https://github.com/owncloud/owncloud-sdk/blob/4116766b9f90e6283d34bbb8eda2baae2ffaf02b/tests/helpers/providerStateHelper.js#L25-L28

In this PR, I have refactored the tests to use state helper functions.

### Related issues
part of #1086

### DIFF
--- a/tests/fileTrashTest.js
+++ b/tests/fileTrashTest.js
@@ -26,7 +26,8 @@ describe('oc.fileTrash', function () {
 
   const {
     givenFolderExists,
-    givenFileExists
+    givenFileExists,
+    givenResourceIsDeleted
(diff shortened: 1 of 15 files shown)


kiranparajuli589 (commented):
lgtm :+1:

individual-it (approved):

PR #1112, opened 2022-07-13
Add range to search response

JanAckermann:
Enhancement: Add range to search result

We've added the range property to the search result, so we can retrieve the found items,
even if the search limit will be exceeded.


lookacat (approved):
Test failes, but LGTM ^^

PR #1134, opened 2022-08-19
Use the resource name WebDAV property

JammingBen:
Needs https://github.com/owncloud/ocis/pull/4412 to be merged.


JammingBen:
Actually this PR is not necessary as we can just use the new property in Web directly.

Issue #1159, opened 2022-10-10
Provider tests against oC10 failing nightly


saw-jan:
Builds: https://drone.owncloud.com/owncloud/owncloud-sdk/3847/9/11, https://drone.owncloud.com/owncloud/owncloud-sdk/3846/9/11

```bash
Verifying a pact between owncloud-sdk and oc-server-pendingOn-ocis

...
  as 'testUser1', a POST request to share a file 'test space and + and #.txt' to user with expiration date set
    returns a response which
      has status code 200 (OK)
      includes headers
        "Content-Type" with value "text/xml; charset=utf-8" (OK)
      has a matching body (FAILED)
  as 'testUser1', a POST request to share a file 'test.txt' to user with expiration date set
    returns a response which
      has status code 200 (OK)
      includes headers
        "Content-Type" with value "text/xml; charset=utf-8" (OK)
      has a matching body (FAILED)
  as 'testUser1', a POST request to share a file '文件.txt' to user with expiration date set
    returns a response which
      has status code 200 (OK)
      includes headers
        "Content-Type" with value "text/xml; charset=utf-8" (OK)
      has a matching body (FAILED)
...
1) Verifying a pact between owncloud-sdk and oc-server-pendingOn-ocis Given the user is recreated And the user is recreated And file exists - as 'testUser1', a POST request to share a file 'test space and + and #.txt' to user with expiration date set returns a response which 
    1.1) has a matching body
           $.ocs.data -> Expected child <displayname_file_owner/> but was missing
           $.ocs.data -> Expected child <displayname_owner/> but was missing
           $.ocs.data -> Expected child <expiration/> but was missing
           $.ocs.data -> Expected child <file_target/> but was missing
           $.ocs.data -> Expected child <id/> but was missing
           $.ocs.data -> Expected child <path/> but was missing
           $.ocs.data -> Expected child <permissions/> but was missing
           $.ocs.data -> Expected child <share_type/> but was missing
           $.ocs.data -> Expected child <stime/> but was missing
           $.ocs.data -> Expected child <uid_file_owner/> but was missing
           $.ocs.data -> Expected child <uid_owner/> but was missing
           $.ocs.meta.message.#text -> Expected 'Expiration date is in the past' to match '(OK)?'
           $.ocs.meta.statuscode.#text -> Expected '100' to be equal to '404'
           $.ocs.meta.status.#text -> Expected 'ok' to be equal to 'failure'
[...]

PR #1160, opened 2022-10-10
[tests-only][full-ci]Added function to get expiration date instead of constant expiration date

SagarGi:
### Description
Previously a constant expiration date was used and the nightly was failing since the expiration date is gone date than the current date. This PR adds a function to that gets the expiration date (+10 days with current date).

### Related NIghtly Failed Issue
https://github.com/owncloud/owncloud-sdk/issues/1159

### DIFF
--- a/tests/helpers/sharingHelper.js
+++ b/tests/helpers/sharingHelper.js
@@ -130,11 +130,26 @@ const getShareIdToken = (resource, resourceType) => {
   return { shareId, shareToken }
 }
 
+/**
+ *
+ *
+ * @returns {String} expiration date (10+ days from current Date)
+ */
+const generateExpirationDate = () => {
+  const currentDate = new Date()
+  const additionalDaysToAdd = 10
+  // adding +10 days to the current data
+  currentDate.setDate(currentDate.getDate() + additionalDaysToAdd)
+  // extracting date in string like '2022-10-10' from currentDate which is in format like '2022-10-10T07:20:33.287Z'
+  return currentDate.toISOString().slice(0, 10)
+}
+
 module.exports = {
   shareResource,
   getShareInfoByPath,
   createFolderInLastPublicShare,
   createFileInLastPublicShare,
   toFormUrlEncoded,
-  getShareIdToken
+  getShareIdToken,
+  generateExpirationDate
 }
--- a/tests/sharingTest.js
+++ b/tests/sharingTest.js
@@ -3,7 +3,8 @@ import { MatchersV3, XmlBuilder } from '@pact-foundation/pact/v3'
 const { fromProviderState } = MatchersV3
 const {
   toFormUrlEncoded,
-  getShareIdToken
+  getShareIdToken,
+  generateExpirationDate
 } = require('./helpers/sharingHelper')
 
 const {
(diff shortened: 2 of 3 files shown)


tests/helpers/sharingHelper.js
```diff
@@ -130,11 +130,23 @@ const getShareIdToken = (resource, resourceType) => {
   return { shareId, shareToken }
 }
 
+/**
+ *
+ *
+ * @returns {String} expiration data (10+ days from current Date)
+ */
+const getExpirationDate = () => {
```
saw-jan, line 138:
```suggestion
const generateExpirationDate = () => {
```

tests/helpers/sharingHelper.js
```diff
@@ -130,11 +130,23 @@ const getShareIdToken = (resource, resourceType) => {
...
+  const currentDate = new Date()
+  currentDate.setDate(currentDate.getDate() + 10)
+  return currentDate.toISOString().slice(0, 10)
```
phil-davis, line 141: I suppose the date is something like `2022-10-20` which happens to be a 10-char-long string.
Future readers (and me) are going to wonder which "10" (or "10s") to change if the date offset needs to be different in the future. A comment will help to explain that `slice(0, 10)` is nothing to do with the `+ 10` in the line above.

SagarGi: @phil-davis sure thanks for reporting !!

tests/sharingTest.js
```diff
@@ -883,7 +883,7 @@ describe('Main: Currently testing file/folder sharing,', function () {
               ...validAuthHeaders,
               ...applicationFormUrlEncoded
             },
-            body: 'shareType=0&shareWith=' + sharee + '&path=' + fileEncoded + '&expireDate=' + expirationDate
+            body: 'shareType=0&shareWith=' + sharee + '&path=' + fileEncoded + '&expireDate=' + getExpirationDate()
```
saw-jan, line 886: call the date generator function somewhere above and use it everywhere below:
```js
const expirationDate = generateExpirationDate()
```
This way we will use same expirationDate for request and response.
```suggestion
            body: 'shareType=0&shareWith=' + sharee + '&path=' + fileEncoded + '&expireDate=' + expirationDate
```

tests/helpers/sharingHelper.js
```diff
@@ -130,11 +130,26 @@ const getShareIdToken = (resource, resourceType) => {
...
+ *
+ *
+ * @returns {String} expiration date (10+ days from current Date)
+ */
+const generateExpirationDate = () => {
+  const currentDate = new Date()
+  const additionalDaysToAdd = 10
+  // adding +10 days to the current data
+  currentDate.setDate(currentDate.getDate() + additionalDaysToAdd)
+  // extracting date in string like '2022-10-10' from currentDate which is in format like '2022-10-10T07:20:33.287Z' with slice
```
saw-jan, line 143: Also, what if the dates are single digit? Will there be two digits format (e.g.: `01`)
```suggestion
  // extracting date in string like '2022-10-10' from currentDate which is in format like '2022-10-10T07:20:33.287Z'
```

SagarGi: yes new Date always give 2 digit format like `20XX-XX-XX`

saw-jan (commented):
LGTM

phil-davis (approved):

PR #1201, opened 2023-01-16
Possibility to pass additional context for the search endpoint and _sendDavReport

elizavetaRa:
- Enables passing of additional context for search and the used by it `_sendDavReport` function.
- In case of path set in the context, passes it to the REPORT request

Motivation:  
- CERNBox sdk search can only search inside the given repository, which needs to be passed to `_sendDavReport`
- Overall nice-to-have for ownCloud(?): the possibility to search inside the given repository in addition to all files

### DIFF
--- a/src/fileManagement.js
+++ b/src/fileManagement.js
@@ -333,10 +333,11 @@ class Files {
    * @param   {string} pattern        pattern to be searched for
    * @param   {number} limit          maximum number of results
    * @param   {string[]} properties   list of DAV properties which are expected in the response
+   * @param   {Object} context        search context
    * @returns {Promise.Object}        boolean: whether the operation was successful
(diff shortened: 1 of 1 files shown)


elizavetaRa:
ping @kulmann @pascalwengerter

src/fileManagement.js
```diff
@@ -447,13 +448,14 @@ class Files {
     return this._sendDavReport(body)
   }
 
-  _sendDavReport (body, plainResult = false) {
+  _sendDavReport (body, plainResult = false, context) {
```
kulmann, line 451: Please give the `context` a default, i.e. `context = {}`.

kulmann (changes requested):

kulmann (approved):
👍

PR #1204, opened 2023-01-26
Allow removing expiration dates from space shares

JammingBen:
`optionalParams.expireDate` is empty when it's being removed from a share. Hence we need to check for `undefined` explicitly.

Needed for https://github.com/owncloud/web/pull/8320.


kulmann (approved):
👍

PR #1235, opened 2023-07-10
[full-ci] [tests-only] Bump ocis commit id to latest

grgprarup:
Part of https://github.com/owncloud/QA/issues/815


phil-davis:
`pactflow_token` was removed from drone in the recent changes.
I have added it to the drone repo again. It is a read-only token to the "pact" data, which can be public anyway. So the value of the token does not have to be secure.

phil-davis (approved):

PR #1245, opened 2023-09-25
Expose statusCode in OCSError

JanAckermann:
Part of:
- https://github.com/owncloud/web/issues/9726

### DIFF
--- a/src/OCSError.js
+++ b/src/OCSError.js
@@ -2,14 +2,16 @@
  * @class OCSError
  * @classdesc A error class for OCS errors
  * @param   {string}  message   message to be added to the OCSError
+ * @param   {number}  statusCode  statusCode to be added to the OCSError
  * @param   {object}  response  http response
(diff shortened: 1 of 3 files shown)


src/OCSError.js
```diff
@@ -2,14 +2,16 @@
...
  * @param   {string}  message   message to be added to the OCSError
+ * @param   {number}  statusCode  statusCode to be added to the OCSError
  * @param   {object}  response  http response
  * @param   {object}  ocs       ocs additional data
 
  */
 
 class OCSError extends Error {
-  constructor (message, response, ocs = null) {
+  constructor (message, statusCode, response, ocs = null) {
```
dschmidt, line 12: Is this class public/exported? Then this is a breaking change and needs a major version bump 🙈

dschmidt (approved):

src/OCSError.js
JanAckermann: No

PR #1247, opened 2023-09-29
[tests-only] bump latest ocis commit id

PrajwolAmatya:
Part of: https://github.com/owncloud/QA/issues/827


SwikritiT (approved):
lgtm if ci is also happy

PR #1273, opened 2024-04-19
docs: add deprecation notice to readme

JammingBen:
Adds a deprecation notice to the `README.md` and prepares the (last and final) release of this package to make this notice available on npmjs.


kulmann (approved):
🥳

